PR #16, opened 2021-05-10
Write forcing

DirkEilander:
- set output chunksize when writing forcing to {time:1, y:-1, x: -1} for fast reading by wflow.jl
- add more options to control chunksize when processing
- fix linux pipeline


hboisgon (commented):
Nice edits!

Some suggestions for the write_forcing method (for later requires core changes):
- add option for rounding the forcing values (with user defined significant digits 1, 2...)
- add option to change the time chunks (lighter forcing file and a bit slower wflow)
- add option for filename

Issue #21, opened 2021-05-12
Add Soilgrids v0.2 data


hboisgon:
Update soilgrids workflow to also work with the new version of Soilgrids.

hboisgon:
There is a new version (2.0) of SoilGrids available: https://soilgrids.org/

download global datasets
include in wflow build

Compared to the 2017 version the data is organized a bit different. Average value per layer is provided, instead of a value that is valid for a certain depth.

laurenebouaziz:
Hi @hboisgon and @verseve , I have made changes in the soilgrids2 branch and wanted to explain here more in detail what I did. Could (one of) you please let me know if additional changes are needed?

Steps undertaken to include soilgrids2 (https://www.isric.org/explore/soilgrids/faq-soilgrids) and adapt the workflow:

1. data download of the complete maps in VRT format through WebDAV (projection in Homolosine projection)
https://files.isric.org/soilgrids/latest/data/ and stored on p drive 

Notes on new maps:
- maps of clay, sand, silt and ph have a slightly different extent than bdod and soc:
bdod and soc: shape 58034, 15924**3** and bounds [-19949000.0, -6147500.0, 19861750.0, 8361000.0]
versus other maps 58034, 15924**6** and bounds [-19949750.0, -6147500.0, 19861750.0, 8361000.0]
There seems to be three columns missing in the bdod and soc maps compared to the others.

I contacted Laura Poggio from soilgrids  and she replied: "_we will look into this and try to make them homogeneous as soon as possible. In the meantime, according to your work-flow I can suggest potential solution. The fastest (computationally) would be to use gdal to change the extent of the files to make them the same._ "

- depth to bedrock is lacking in the 2020 version. Laura indicated that they are working on it and that the new maps will become available at the end of this year. She did not reply if she expected large changes for this variable.

- soil taxonomy is also lacking, Laura indicated that they do not plan to provide this map anymore. 

Based on these differences I did the following preprocessing steps:

2. Preprocess 
- clip sand, clay, silt and ph to the extent of soc and bdod (I first tried a gdalwarp, but this did not work)
- soilthickness map -- use the map from 2017 version -- gdalwarp to homolosine projection and extent of other maps and convert to vrt format (but I haven't done this nicely with separate tiles - instead it is just a single tile, this could be improved. My main aim now was to have all files as vrt to read them from the yaml with a similar *_mean.vrt extension.
- taxonomy map -- same procedure as soilthickness map

3. yml file:
[...]

PR #31, opened 2021-05-18
adapted workflow for soilgrids2

laurenebouaziz:
see #21 for detailed description of changes.


verseve:
I have focused on the proposed calculations for Soilgrids 2.0 and implementation is correct. One question: for the c parameter, is it perhaps better to use a weighted average here?:
https://github.com/Deltares/hydromt_wflow/blob/soilgrids2/hydromt_wflow/workflows/soilgrids.py#L435

My suggestion is to also add the calculation approaches for the different Soilgrids versions to the documentation.

laurenebouaziz:
Willem, thanks for checking the code. not sure if I understand the question correctly, isn't the weighted average computed in lines 441-454? 
good point to document better the different calculation approaches.

verseve:
Ah, yes you're right! I was looking at the wrong soil depth array (midpoint).

hydromt_wflow/workflows/soilgrids.py
```diff
@@ -165,21 +155,19 @@ def pore_size_distrution_index_layers(ds, thetas):
...
-            ds["clyppt_sl" + str(l)],
-            dask="parallelized",
-            output_dtypes=[float],
-        )
-        da.name = "sl" + str(l)
-        da = da.raster.interpolate_na()
-        da_lst.append(da)
-    ds = xr.merge(da_lst)
-    return ds
+    da = xr.apply_ufunc(
```
hboisgon, line 158: Use da for DataArray and ds for Dataset. To ease comprehension for the developper :)

hboisgon (commented):
Nice additions to handle soilgrids 2020! 
In the end it's not that many "if soilgrids =", only during resampling over the different soil layers which is to be expected. Good that we can keep both options.
There's one comment somewhere on conventions on the use of da vs ds but minor :)

One last thing, I didn't see the yml file you used to read soilgrids data so not sure if this is relevant but there was no way to concat over soil layers direclty when reading soilgrids data (passing kwargs for the concat and concat_dim of this function https://github.com/Deltares/hydromt/blob/a3f5dc16248e4335ef89df749d87e2b72985c97c/hydromt/io.py#L70 )?

laurenebouaziz:
thanks! 
I have adapted whenever applicable da to ds.
We had checked with Dirk, but it was not possible to right away use concat_dim to merge the layers, as far as I understood. 

Here is the yaml for just soilgrids_2020 to be added to the artifacts
[soilgrids_2020.zip](https://github.com/Deltares/hydromt_wflow/files/6632023/soilgrids_2020.zip)

laurenebouaziz:
If I see it correctly, tests are failing on the wflow_soil.map as this map was completely changed.
@hboisgon , do we need to change the tests or should we change wflow_soil to soil_texture  and keep the previous map?

hboisgon:
The tests are here to check that when changing the code you do not get unexpected changes or results downstream. 

Here you changed the computation method for wflow_soil.map so that the changes for this map is expected :) as this is something we want to keep from now on, then you need to update the examples used for the tests as well (here this means update the staticmaps of all models: wflow_piave_subbasin, wflow_piave_clip and wflow_piave_sediment).

Issue #40, opened 2021-07-23
Implement results and read_results method


hboisgon:
Adding a results object and the corresponding read_results method.

To read the results and know where and what is stored in the output nc and csv file, the corresponding TOML file and settings are needed. This means **one TOML file per wflow run**. The path and names of the files are also read from there.

The results objects would be a dictionnary containing **(edit after comments from Dirk)**:

- **a "output" xr.Dataset** for the gridded netcdf file (output section of the TOML)
- **a "netcdf" xr.Dataset** for the netcdf point timeseries file (neycdf section of the TOML)
- **different GeoDataArrays for the csv file** , one per column (csv section and csv.column sections of the TOML). The xy coordinates are the coordinates of the station or of the centroid of the subcatch/area. The variable name in the GeoDataArray corresponds to the csv header attribute.

DirkEilander:
Good proposal @hboisgon ! We could also start saving results as a dictionary of Datasets and/or DataArrays. I would prefer that over a dictionary with dictionaries of DataArrays. This is closer to how the wflow data is saved and will also make some postprocessing steps easier, e.g.: to do zonal statistics for all variables in a Dataset can be done in one step instead of looping over the DataArrays. This would require a change in the `set_results` and `test_model_api`

hboisgon:
I think a dictionnary of datasets and datarrays is better indeed @DirkEilander !
Would that be a problem for sphincs if I adapt the `set_results` method? 
Or is it better to add an argument to the function that splits the dataset into dataarray if true and keeps the dataset if false?

DirkEilander:
An argument whether to split the Dataset into DataArray sounds good. I think by default it should not split (more intuitive) and keep the Dataset. Will make this work on the SFINCS side after your implementation.

hboisgon:
Update of the set_results method of the model API is done here: https://github.com/Deltares/hydromt/pull/58

Issue #17, opened 2021-05-10
linux test fail on soilgrids map with large differences


DirkEilander:
My guess is that these differences originate from the `raster.interpolate_na` method. Perhaps the fill values are not parsed correctly? Or the grid coordinates have rounding errors causing differences in the nearest interpolation at cells with missing values.

see https://github.com/Deltares/hydromt_wflow/runs/2544491332 
relates to #4 

for wflow: 

```
AssertionError: 14 invalid maps: {
  'SoilThickness': 'diff: -14.7917 (3 cells: [(12.066667, 46.550000), (11.850000, 46.433333), (12.333333, 46.133333)])', 
  'SoilMinThickness': 'diff: -14.7917 (3 cells: [(12.066667, 46.550000), (11.850000, 46.433333), (12.333333, 46.133333)])', 
  'KsatVer': 'diff: 1.7397 (12 cells: [(12.066667, 46.550000), (12.083333, 46.550000), (11.833333, 46.516667), (11.850000, 46.433333), (12.016667, 46.400000), (12.066667, 46.166667), (12.333333, 46.133333), (12.350000, 46.116667), (12.316667, 46.100000), (12.333333, 46.100000), (12.216667, 45.983333), (12.233333, 45.983333)])', 
  'KsatVer_0.0cm': 'diff: 1.7397 (12 cells: [(12.066667, 46.550000), (12.083333, 46.550000), (11.833333, 46.516667), (11.850000, 46.433333), (12.016667, 46.400000), (12.066667, 46.166667), (12.333333, 46.133333), (12.350000, 46.116667), (12.316667, 46.100000), (12.333333, 46.100000), (12.216667, 45.983333), (12.233333, 45.983333)])', 
  'KsatVer_5.0cm': 'diff: 1.2059 (11 cells: [(12.066667, 46.550000), (12.083333, 46.550000), (11.833333, 46.516667), (11.850000, 46.433333), (12.016667, 46.400000), (12.066667, 46.166667), (12.333333, 46.133333), (12.350000, 46.116667), (12.333333, 46.100000), (12.216667, 45.983333), (12.233333, 45.983333)])', 
[...]

PR #42, opened 2021-07-28
Results

hboisgon:
Implementation of the results obeject and set_results method for wflow.

For the discussion and info, see issue: #40 

Works ideally with the new hydromt release v0.4.2


laurenebouaziz:
hi Hélène, 

I have not had a chance yet to look at the code in detail unfortunately. However, I tested the exemple that you made for my testcase, and encountered the following issue:

In my toml, I am using the same header for several gauges files (which all contain different ids), as shown below:

[csv]
path = "output.csv"

[[csv.column]]
header = "Q"
map = "gauges_S01"
parameter = "lateral.river.q_av"

[[csv.column]]
header = "Q"
map = "gauges_S02"
parameter = "lateral.river.q_av"

[[csv.column]]
header = "Q"
map = "gauges_S03"
parameter = "lateral.river.q_av"

[[csv.column]]
header = "Q"
map = "gauges_S04"
parameter = "lateral.river.q_av"

[[csv.column]]
header = "Q"
map = "gauges_S05"
parameter = "lateral.river.q_av"

[[csv.column]]
header = "Q"
map = "gauges_S06"
parameter = "lateral.river.q_av"

[[csv.column]]
header = "Q"
map = "gauges_grdc"
parameter = "lateral.river.q_av"


when I then run:

# Example of accessing the results
runs[mainrun]['mod'].results

it says:
Replacing result: Q
Replacing result: Q
Replacing result: Q
Replacing result: Q
Replacing result: Q
Replacing result: Q

so I think it does not append all the different Q_station, but it replaces and only keeps the last one. So then when I try to plot results for the Q station in file S01, it only has the Q of the GRDC files. 

I did not have time to go through the code to see where I can change this though. 

Other than that, it is really easy to use and set up!!! so I think it is of great value :)

hboisgon:
Thanks @laurenebouaziz 

In order to not overwrite, I now use the {header}_{map} attributes of [[csv.column]] to create the names in the results dict.
The other solution would be to merge the different series together but this would require some checks on the values inside the gauges_S0* maps to be sure that they dont use the same index for different gauges (ex gauges numbered from 1 to 10 in all the maps).

hboisgon:
The tests were still failing probably because of an update of gdal/pyproj. The staticmaps that are different are all linked to soil parameters for which 4 cells have different values (always above a NaN zone in the soilgrids data). So I suspect something similar than in #17

The (relative) difference is small enough so for now the staticmaps of the piave examples models were updated.

hydromt_wflow/wflow.py
```diff
@@ -1606,11 +1631,175 @@ def write_states(self):
...
+            for col in self.get_config("csv.column"):
+                header = col["header"]
+                # Column based on map
+                if "map" in col.keys():
+                    # Read the corresponding map and derive the different locations
+                    # The centroid of the geometry is used as coordinates for the timeseries
+                    map_name = self.get_config(f"input.{col['map']}")
+                    da = self.staticmaps[map_name]
+                    gdf = da.raster.vectorize()
+                    gdf.geometry = gdf.geometry.centroid
```
DirkEilander, line 1669: I suggest to use the basin outlet or otherwise representative_point as the centroid might be located outside the basin.
https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoSeries.representative_point.html

hydromt_wflow/wflow.py
```diff
@@ -1606,11 +1631,175 @@ def write_states(self):
...
+                        csv_fn, name=f'{header}_{col["map"]}', usecols=usecols
+                    )
+                    da = GeoDataArray.from_gdf(gdf, da_ts, index_dim="index")
+                # Column based on xy coordinates or reducer for the full model domain domain
+                else:
+                    # Read the timeseries
+                    usecols = [0]
+                    usecols = np.append(usecols, np.arange(count, count + 1))
+                    count += 1
+                    try:
```
DirkEilander, line 1688: Why do we need the try/ except clause here? Is this something we could identify before trying open_timeseries_from_table or somethin we could adapt within that method?

hydromt_wflow/wflow.py
```diff
@@ -1606,11 +1631,175 @@ def write_states(self):
...
+        # Read scalar netcdf (netcdf section)
+        ncs_fn = self.get_config("netcdf.path", abs_path=True)
+        if ncs_fn is not None and isfile(ncs_fn):
+            self.logger.info(f"Read results from {ncs_fn}")
+            ds = xr.open_dataset(ncs_fn, chunks={"time": 30})
+            self.set_results(ds, name="netcdf")
+
+        # Read csv timeseries (csv section)
+        csv_fn = self.get_config("csv.path", abs_path=True)
+        if csv_fn is not None and isfile(csv_fn):
```
DirkEilander, line 1656: Would it be possible to make the csv result parsing a separate method outside the model class in e.g.: wflow_utils.py script (I did something similar for SFINCS)? That makes unit testing easier and also allows the user to use this method outside the model class.

DirkEilander (commented):
@hboisgon: Nice addition to the wflow plugin! I added some small suggestions up for discussion.

hydromt_wflow/wflow.py
hboisgon: Good suggestion to use the basin outlet! I'll add this in the issue.

hydromt_wflow/wflow.py
hboisgon: I think the problem is when non numeric column heads are used in the csv file, then the method cannot find a numeric index. I used the try/except to avoid repeating many lines for the different wflow csv output options. 
I also see the non numeric index issue when trying to open observed csv timeseries so maybe something we could still adapt in that method?

hydromt_wflow/wflow.py
hboisgon: Good idea to add a wflow_utils.py script for this. I also started to do something similar for RIBASIM and delwaq :)

PR #43, opened 2021-07-30
add lines glacier toml

laurenebouaziz:
default add lines for glacier parameters and extent


hboisgon (commented):
Nice addition and in line with what we do for reservoirs/lakes :)

You also need to update the toml files in the different examples cases. I think this is why the test fails.

laurenebouaziz:
ah ok, will do! I was already wondering why it failed..

PR #45, opened 2021-08-06
Flex classification

laurenebouaziz:
hi, I added a workflow to calculate the percentages of each hydrological response unit based on hand and slope, which is data needed to run the flextopo model.  
@hboisgon , could you have a look at the implementation (no rush, after your holidays is fine!)?

### DIFF
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -89,7 +89,6 @@ class WflowModel(Model):
         "percentH": "wflow_percentH",
         "percentP": "wflow_percentP",
         "percentW": "wflow_percentW",
-
     }
     _FOLDERS = [
         "staticgeoms",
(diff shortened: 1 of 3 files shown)


laurenebouaziz:
I think this should probably be done in a separate wflow flextopo class?

hboisgon:
Hi Laurène, nice that you are strating to work on flextopo!
I think it is better to define a new model for wflow_flextopo, wflow is actually more wflow_sbm. Especially if you need different input for wflow_flextopo.

You can have a look at wflow_sediment for an example. In principle you can define wflow_flextopo as a subclass of WflowModel instead of the model API. In that case every function that is completely the same can still be called without needing to re-define it (exemple setup_basemaps for wflow_sediments). If some functions are a little different, you can first call the main super function and then afterwards do flextopo specific operations (exemple setup_lakes). Or even re-use the super function but with slightly different function parameters (exemple setup_lulcmaps or setup_gauges).

Hope this helps you get nicely started :)

laurenebouaziz:
thanks @hboisgon this will definitely be useful to get started :) !

PR #56, opened 2021-10-08
implement methods to create maps for wflow sbm + gw

laurenebouaziz:
added maps for:
- surface area per cell
- aquifer thickness
- conductivity
- specific yield (uniform)
- drain conductance
- drain elevation
- infiltration and exfiltration conductance
- river bottom elevation (other approach than for RiverZ map)

to do: 
- move surface area calc ?
- cell_area - wetted_area check 
- changelog
- yml file extra entry for aquifer thickness

### DIFF
--- a/examples/wflow_sbmgw_build.ini
+++ b/examples/wflow_sbmgw_build.ini
@@ -1,0 +1,98 @@
+[global]
+data_libs       = []            # add optional paths to data yml files
+
+[setup_config]                  # options parsed to wflow ini file <section>.<option>
+starttime = 2010-01-01T00:00:00
+endtime = 2010-03-31T00:00:00
+timestepsecs = 86400
+input.path_forcing = inmaps-era5-2010.nc
+
+[setup_basemaps]
+hydrography_fn     = merit_hydro   # source hydrography data {merit_hydro, merit_hydro_1k}
+basin_index_fn  = merit_hydro_index # source of basin index corresponding to hydrography_fn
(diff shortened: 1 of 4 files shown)


verseve:
>     * For the 1D+2D local inertial based routing we will also need to derive a river depth / river bed elevation map. Would be good to align this both in hydromt_wflow and wflow.jl.

For the 1D local inertial routing I have been using the `RiverZ` map so far (as part of wflow_sbm, without flooding).  @DirkEilander: I guess the river bottom elevation mentioned here is not the approach you are working on?

laurenebouaziz:
The "river_bottom_elevation" map that I added for the groundwater module is based on a very simple relation relating river width to river depth (taken from the paper of Neal et al. 2012). Pretty sure that this is different than what you are working on now @DirkEilander.

The RiverZ map is based on differences in the DEM from the high resolution elevation to the model resolution. This means that in areas with steep slopes, you get a very high depth (as in the Ardennes for the Meuse sometimes more than 100m) and in the flatter parts you get very low river depths, which might not be very realistic.

The simple formulation of river_bottom_elevation leads to a very different pattern, with depths of about 6m at Borgharen and about 1m in the tributaries.

See the two maps below (note the differences in scale!)
(I am testing both maps for the local inertial routing in wflow, had an error previously while running the one with river_bottom_elevation, not sure if the connection to the p-drive failed, so work in progress)

river_depth_river_bottom_elevation
river_depth_RiverZ

Issue #57, opened 2021-10-18
lake_b parameter


verseve:
The calculation of lake_b parameter seems wrong:

```python
            gdf_org["Lake_b"] = (
                gdf_org["LakeAvgOut"].values / gdf_org["LakeAvgLevel"].values
            ) ** (2)
```

should be:

```python
            gdf_org["Lake_b"] = gdf_org["LakeAvgOut"].values / (gdf_org["LakeAvgLevel"].values) ** (2)
```

PR #58, opened 2021-10-19
Lake_b calculation

hboisgon:
Bugfix for calculation of lake_b parameter in setup_lakes, see #57


DirkEilander:
Thanks for fixing @hboisgon! Could you also update the changelog?

PR #63, opened 2021-11-17
river bathymetry

### LINKED ISSUE #24: river width improvements
Several ideas about river width improvements from the old gitlab repos:
> general improvements https://gitlab.com/deltares/wflow/hydromt/-/issues/58
> add user supplied width based on a shapefile/ GRWL, see https://gitlab.com/deltares/wflow/hydromt/-/issues/131

DirkEilander:
- adds new workflows.river.river_bathymetry method used in WflowModel.setup_rivers
- adds new WflowModel.setup_floodplains method
- fixes #24 

TODO
- [x] update tests
- [x] bump hydromt core and pyflwdir version after upstream releases
- [x] add FloodplainZ map to toml file


DirkEilander:
This PR is ready to merge after the open TODOs. 

It can be tested by installing the main branch of pyflwdir or release v0.5.3 on pypi (conda-forge will follow asap) and the rivers branch of hydromt, see https://github.com/Deltares/hydromt/pull/79 (will be released asap)

@verseve Could you already have a look at let me know what you think. I also couldn't find how the **FloodplainZ** should be referenced in the wflow toml file.

verseve:
Yes, will look at this today, including setting up an updated Wflow test model for the Moselle (part of the Wflow test bank). The FloodplainZ is now mapped in the TOML file through `input.altitude`, will change this to `input.lateral.land.elevation` since it is directly part of this model component.

verseve:
Did generate a wflow model for the Moselle catchment as follows:
hydromt build wflow "./" "{'subbasin': [7.58696, 50.36290], 'strord': 8}" -r 0.00833 -i wflow_build.ini --dd

When running the model I noticed a couple of grid cells with relative high water depths. For example this is the case for index (row, column)  `[216, 77]`, a land cell (river cells are located at `[215, 77]` and `[216, 78]`). `FloodplainZ[216, 77]` is 227.505 m, while surrounding cells in D4 are higher (a pit). Both surrounding river cells have a `FloodplainZ` of 232.671, `RiverZ` is 213. I have also compared this to the previous generated `wflow_dem_d4` map, and there the elevation of `[216, 77]` is equal to the two surrounding river cells (227.505 m). 

Not sure if this is an issue?

verseve:
With the current implementation for 1D-2D local inertial flow, there is a mismatch between `RiverZ`, `RiverDepth` and `FloodplainZ`. For the Wflow model type `sbm` in combination with lateral subsurface flow (kinematic wave) this does not cause any problems, because there is no exchange between subsurface flow and river, and for floodplain flow Wflow checks on `RiverDepth`. 

However, for the model type `sbm` with `GroundwaterFlow` (also related to the work of @laurenebouaziz for the Meuse basin) this is not the case, `RiverZ`, `RiverDepth` and `FloodplainZ` should actually match. Any thoughts on this @DirkEilander  and @laurenebouaziz ?

DirkEilander:
> When running the model I noticed a couple of grid cells with relative high water depths. For example this is the case for index (row, column) [216, 77], a land cell (river cells are located at [215, 77] and [216, 78]). FloodplainZ[216, 77] is 227.505 m, while surrounding cells in D4 are higher (a pit).
> 
> Not sure if this is an issue?

This issue has been solved by added an additional preprocess step before deriving the D4 flow direction which are needed to hydrologically condition the DEM in D4.

DirkEilander:
@verseve @laurenebouaziz I've now implemented the following workflow based on our discussions yesterday:

- In `[setup_rivers]` we derive a `RiverZ` map which contains the elevation at outlet subgrid pixels for the entire domain. This map is used to set the `input.lateral.river.bankfull_elevation` in the toml file for 1D simulations.
- In `[setup_floodplain]` , which is only required for 1D/2D simulations, either the `RiverZ` or `wflow_dem` (default) map is hydrologically conditioned to D4 flow direction for the enitre domain and additionally to D8 flow direction for the river cells. The user can choose which one. This map is saved as `FloodplainZ` and is used to (re)set both the `input.lateral.river.bankfull_elevation` and `input.lateral.land.elevation` in the toml file.

left to do:

- [ ] move hydromt_wflow.worlflows.river.adjust_dem to hydromt.flw.adjust_dem

laurenebouaziz:
sounds in line with what we discussed yesterday! curious to see what the effect is of both maps for the floodplain elevation.

verseve:
Sounds good @DirkEilander.
I did a bit more testing, not very extensive (mostly based on 1 station), for the Moselle catchment. Coupe of comments:
- Most important observation: the map `N` (Manning's roughness for land surface) works fine in combination with the kinematic wave, but for the local inertial approach there is too much dampening. When using the default value for N (from Wflow code) results improve significantly. I think for this concept it is necessary to provide different (more realistic? if I remember correctly the N values for the kinematic wave overland flow are quite high) values. 
- For both `z_subgrid` and `z_avg` too much retention of water (because of possible overestimation of the floodplain storage) does not seem to happen. With `z_subgrid` the system is more reactive (faster rising hydrograph, higher peaks) compared to `z_avg`.
- Computational times can be large. This is of course also partly caused by very short channel lengths, the minimum river length for the Moselle catchment is for example 76 m.

laurenebouaziz:
interesting @verseve and do you see a substantial reduction of the peak flows compared to the kinematic or local inertial without inundation approach for the Moselle?

verseve:
> interesting @verseve and do you see a substantial reduction of the peak flows compared to the kinematic or local inertial without inundation approach for the Moselle?

Yes, this is for sure possible with this concept (peak flow reduction). My testing period was quite short (4 months), it also depends on the `FloodplainZ` map used (average elevation gives more dampening) in combination with Manning's n used (now just one value). As an example, for a measured peak of 674 m3/s, full kinematic wave gives 804 m3/s and for 1D/2D with `z_avg` this is 739 m3/s.

DirkEilander:
Thanks for the feedback @verseve 

- I'm not surprised that z_subgrid is more reactive as I would  expect there is a steeper (more realistic?) slope for the river cells compared to z_avg
- Do you have a suggestion how to improve this? Manning values are derived from these tables and are based on landuse classes. Would be interesting to compare different data sources.  https://github.com/Deltares/hydromt_wflow/tree/main/hydromt_wflow/data/lulc 
- There's not much we con do about this in the current setup. A good double check to see how important the  lengths really are would be to use a spatial uniform length based on the average of all river cells, just to compare run times.

verseve:
>     * For both `z_subgrid` and `z_avg` too much retention of water (because of possible overestimation of the floodplain storage) does not seem to happen. With `z_subgrid` the system is more reactive (faster rising hydrograph, higher peaks) compared to `z_avg`.

I did notice that results of above mentioned test with `z_subgrid` and `z_avg` were based on an incorrect `RiverDepth` setting (fixed to a high value so water stays in the river). Re-running the tests with the correct `RiverDepth` map shows quite different results, with `z_avg` the hydrograph is much delayed and almost non-responsive, with `z_subgrid` it is more responsive but with too much delay. Besides a possible underestimation of river width and depth (not very likely) I think:

-  the combination of D8 for the river and D4 for overland flow is most likely not ideal, with diagonal river directions the floodplain water can't follow the river, with (possibly) loosing/delaying too much water. With `z_subgrid` the issue seems mostly too much delay.

Possible solutions/ worth testing:

1.  Use for this concept D4 for overland and river flow. I expect minor changes on Wflow.jl side, not sure for HydroMT; how is this for example implemented for the SFINCS model?
2. Allow for river cells with a floodplain D8, or even D8 for the whole domain (this is for example also implemented in the LeafHYDRO-Flood model).

DirkEilander:
Thanks for the update Willem. 

Before trying new solutions I think it might be worthwhile investigating if we can improve the data. 
- An underestimation of river depth is not unlikely as it is conditioned to a discharge value which might be different from what we simulate. There are multiple options in hydroMT that we can try, see https://github.com/Deltares/hydromt_wflow/blob/1f8a85e7566e533ff2aa7e01b5f7eb7501e054e6/hydromt_wflow/workflows/river.py#L167
- I've also tried to solve the D4 floodplain connectivity by conditioning the elevation to D4 flow directions. Can you find locations within the basin where this is an issue to evaluate if this was done correct and/re need further improvements.
- And not sure if this is already possible/ you did this, but it would also be useful to store additional variables such as overbank flow into the floodplains (and overland flow into the rivers), to analyzing the issue.

verseve:
> * And not sure if this is already possible/ you did this, but it would also be useful to store additional variables such as overbank flow into the floodplains (and overland flow into the rivers), to analyzing the issue.

Yes, I focused a bit more on two GRDC locations. For 6336900, index `[121, 147]` (row, column) that shows too much delay for the falling limb of the hydrograph and for smaller peaks too much delay and dampening. The upstream river cell `[123, 149]` flows to river cell `[122, 148]`. Floodplain flow (because of D4) goes to land cell `[123, 148]` , and shows (as an example) for 2000-02-21 a water depth of 0.66 m (water depth in `[123, 149]` is 2.20 m with a `RiverDepth` of 1.46 m). A similar situation occurs at another upstream river cell `[125, 153]` and I expect there are many more similar locations where this occurs.
Land cell `[123, 148]` has the same `FloodplainZ` value as river cell `[123, 149]` (170.2 m, `z_subgrid`). I guess this is caused by the D4 conditioning (instead of applying `RiverZ` (subgrid outlet elevation) to a Wflow land cell)?

For GRDC station 6336800, index `[113,113]` there is also too much delay (rising and falling limb) and too much dampening. A couple of upstream locations (as example) where quite some water is 'lost' to adjacent land cells: `[117, 112]` , `[116, 110] `, `[118, 107]` ... because of diagonal flow directions for river cells. 

> * I've also tried to solve the D4 floodplain connectivity by conditioning the elevation to D4 flow directions. Can you find locations within the basin where this is an issue to evaluate if this was done correct and/re need further improvements.

The D4 conditioning seems to work fine generally. There is only one location `[ 268, 191]` (land cell) that I noticed where water depth is quite high and this is indeed a D4 pit. 

Yes, an underestimation of river depth is of course possible, although not sure how much / enough room there is to solve this issue completely. Also, my impression is, based on the small analysis I did, that this issue still occurs when `RiverDepth` is more accurate and bankfull depth is exceeded: too much water is `lost` to adjacent land cells, because of combining D4 and D8. So, I think it would be good to consider/test the other options mentioned before. 

What do you think, based on these results?

DirkEilander:
I thinks this is ready to merge. Would be good to 1) check if the API and docs are clear and 2) double check technical implementations, equations etc.

examples/wflow_piave_clip/wflow_sbm.toml
```diff
@@ -117,6 +117,7 @@ length = "wflow_riverlength"
 n = "N_River"
 slope = "RiverSlope"
 width = "wflow_riverwidth"
+bankfull_elevation = "RiverZ"
```
verseve, line 120: What about `bankfull_depth` here? 

Here `RiverZ` is now used as `bankfull_elevation`. This is probably for the 1-D local inertial approach for river flow (without the 2-D part)? Also related to the previous comment, when using the 1-D/2-D local inertial approach, is it fine to use `RiverZ` in combination with `FloodplainZ` for overland flow? Would be good to add a bit more detail to the docs.

The toml file is probably not changed when using the local inertial approach for wflow_sbm, right? I think this is fine, it is explained in the Wflow docs:
```toml
[model]
river_routing = "local-inertial"
land_routing = "local-inertial"
```

verseve (changes requested):
This looks good, I've two general comments:
- about the method name `setup_floodplains` (and `FloodplainZ`), and what basis data/methods are used (related to the provided `z_subgrid` and `z_avg` maps for testing the Moselle catchment), and to add this to the docs.
- related to the previous comment, when to use `RiverZ` and `FloodplainZ`, and to add this to the docs.
-

examples/wflow_build.ini
```diff
@@ -13,9 +13,17 @@ basin_index_fn  = merit_hydro_index # source of basin index corresponding to hyd
...
+hydrography_fn   = merit_hydro      # source hydrography data, should correspond to hydrography_fn in setup_basemaps
+river_geom_fn    = rivers_lin2019_v1 # river source data with river width and bankfull discharge
+river_upa        = 30               # minimum upstream area threshold for the river map [km2]
+rivdph_method    = powlaw           # method to estimate depth {'powlaw', 'manning', 'gvf'}
+min_rivdph       = 1                # minimum river depth [m]
+min_rivwth       = 30               # minimum river width [m]
+slope_len        = 2000             # length over which tp calculate river slope [m]
+smooth_len       = 5000             # length over which to smooth river depth and river width [m]
+
+[setup_floodplains]             # adds hydrologically conditioned elevation FloodplainZ
```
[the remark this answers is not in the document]
DirkEilander: @verseve Do you have a suggestion for a better name instead of `FloodplanZ`? What do you think of the following (also curions to hear your thoughts on this naming @laurenebouaziz @hboisgon):

- `wflow_dem` -> `dem_avg`
- `RiverZ` -> `dem_subgrid` This dem is based on the subgrid outlet elevation levels, hydrologically conditioned to D8 flowdirs and not just for the river cells anymore.
- `FloodplainZ` -> `dem_avg_d4` (or `dem_subgrid_d4`) depending on which of the above it s based on. This map has D4 conditioned elevation for land and D8 conditioned for river cells.

It depends on the `elevnt_map`  argument if the FloodplainZ is based on the subgrid or average dem, see
```
      Parameters
      ----------
      elevtn_map: str, optional
          Name of staticmap to base the river+floodplain elevation map on, by default
          "wflow_dem". To base it on the subgrid river elevation map use "RiverZ";
          to base it on avarage cell elevation use "wflow_dem".
```

examples/wflow_piave_clip/wflow_sbm.toml
DirkEilander: `bankfull_depth` is currently not in the deafult toml file. Only if the `river_geom_fn` argument of the `WflowModel.setup_rivers` method is provided the `wflow_riverdepth` is calculated and the `bankfull_depth` map set in the toml file.

The  toml file is not automatically changed to use the local inertial approach. To use the local-inertial method it's best to set this option in the toml file through the `WflowModel.setup_config` method or directly in the toml file.

examples/wflow_build.ini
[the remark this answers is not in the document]
verseve: Thanks for the answers Dirk. I agree with your suggestions, with these names it is much easier to interpret these variables.

For the `setup_floodplain` my suggestion is to use for example `setup_hydrodem`.

hydromt_wflow/wflow.py
```diff
@@ -235,95 +227,190 @@ def setup_basemaps(
...
+            self.set_config("input.lateral.river.bankfull_depth", self._MAPS["rivdph"])
 
         self.logger.debug(f"Adding rivers vector to staticgeoms.")
         self.staticgeoms.pop("rivers", None)  # remove old rivers if in staticgeoms
         self.rivers  # add new rivers to staticgeoms
 
+    def setup_floodplains(self, elevtn_map="wflow_dem"):
+        """This component adds a hydrologically adjusted river+floodplain elevation map
+        for the 1D-2D local inertial based routing.
+
```
laurenebouaziz, line 383: add a note to specify hydrologically adjusted with d4 for land and d8 for rivers cellls in docstring?

hydromt_wflow/workflows/river.py
```diff
@@ -32,31 +27,31 @@ def power_law(x, a, b):
...
     """Returns river maps
 
     The output maps are:\
     - rivmsk : river mask based on upstream area threshold on upstream area\
     - rivlen : river length [m], minimum set to 1/4 cell res\
     - rivslp : smoothed river slope [m/m]\
-    - rivwth_obs : river width at pixel outlet (if in ds)
-    - rivbed : elevation of the river bed based on pixel outlet 
+    - rivzs : elevation of the river bankfull height based on pixel outlet 
+    - rivwth : river width at pixel outlet (if in ds)
```
laurenebouaziz, line 44: is it good to mention that this is the bankfull river width here? (it depends of course on the river_geom_fn)

examples/wflow_build.ini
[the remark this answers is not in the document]
laurenebouaziz: I agree that the suggested names are clear like this

laurenebouaziz:
Really nice work Dirk!
I just made some very minor suggestions and corrected typos as I was reading through the docstrings. 
also added a few minor notes in a rivers branch of hydromt for river_depth.

one other suggestion would be to also output the bankfull discharge as a map in staticmaps (I checked the rivers.geojson but it does not contain the bankfull Q). This would make it possible to compare the values with the wflow runs.

I updated the Meuse model with the 3 different options for river depth and dem_avg / dem_subgrid. I will then for these testcases run the wflow model and update you later on the findings.

laurenebouaziz:
I tried to run the model, but I am getting an error. At first I thought it was perhaps a missing value in the netcdf but it seems more as if the loop exceeds the number of indices:

this is part of the logging with the error:

```
┌ Info: Local inertial approach is used for overlandflow.
│   alpha = 0.7
│   theta = 0.8
│   h_thresh = 0.001
└   froude_limit = true
[ Info: Set lateral.land.n using default value 0.072.
[ Info: Set lateral.land.elevation using NetCDF variable FloodplainZ.
ERROR: BoundsError: attempt to access 51804-element Vector{Float64} at index [51805]
Stacktrace:
  [1] getindex
    @ .\array.jl:801 [inlined]
  [2] set_effective_flowwidth!(we_x::Vector{Float64}, we_y::Vector{Float64}, indices::Wflow.Indices, graph_riv::Graphs.SimpleGraphs.SimpleDiGraph{Int64}, riverwidth::Vector{Float64}, ldd_riv::Vector{Union{Missing, UInt8}}, inds_rev_riv::Vector{Int64})
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\utils.jl:552
  [3] initialize_shallowwater_land(nc::NCDatasets.NCDataset{Nothing}, config::Wflow.Config, inds::Vector{CartesianIndex{2}}; modelsize_2d::Tuple{Int64, Int64}, indices_reverse::Matrix{Int64}, xlength::Vector{Float64}, ylength::Vector{Float64}, riverwidth::Vector{Float64}, graph_riv::Graphs.SimpleGraphs.SimpleDiGraph{Int64}, ldd_riv::Vector{Union{Missing, UInt8}}, inds_riv::Vector{CartesianIndex{2}}, river::Vector{Bool}, Δt::Dates.Second)
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\flow.jl:863
  [4] initialize_sbm_model(config::Wflow.Config)
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\sbm_model.jl:207
  [5] run(config::Wflow.Config)
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:176
  [6] (::Wflow.var"#178#179"{IOStream, Bool, Wflow.Config})()
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:155
  [7] with_logstate(f::Function, logstate::Any)
    @ Base.CoreLogging .\logging.jl:491
  [8] with_logger(f::Function, logger::LoggingExtras.TeeLogger{Tuple{LoggingExtras.EarlyFilteredLogger{Logging.ConsoleLogger, Wflow.var"#172#175"}, LoggingExtras.EarlyFilteredLogger{LoggingExtras.MinLevelLogger{TerminalLoggers.TerminalLogger, Base.CoreLogging.LogLevel}, Wflow.var"#171#174"}}})
    @ Base.CoreLogging .\logging.jl:603
  [9] run(tomlpath::String; silent::Nothing)
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:151
 [10] run
    @ c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:144 [inlined]
 [11] run()
    @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:248
 [12] top-level scope
    @ none:1
```

I am running this toml:
`p:\11205237-grade\wflow\wflow_meuse_julia\wflow_sbm_meuse_adapted_river_powlaw_subgrid\run_default\wflow_sbm_eobs_1980_2020.toml`

with the following code:
`julia  -t 4 --project=c:\Users\bouaziz\.julia\dev\Wflow -e "using Wflow; Wflow.run()" "wflow_sbm_meuse_adapted_river_powlaw_subgrid\run_default\wflow_sbm_eobs_1980_2020.toml"`

I have the same error for the model case with river depth gvf method in combination with dem_avg. 

@verseve did you encounter something similar while testing the Moselle?

verseve:
> I tried to run the model, but I am getting an error. At first I thought it was perhaps a missing value in the netcdf but it seems more as if the loop exceeds the number of indices:

Yes, indeed something goes wrong with the indices in the function `set_effective_flowwidth!`. Haven't seen this with testing the Moselle, will look into your issue with the Meuse model.

DirkEilander:
That's a quick fix @verseve ! So I understand it does not require additional changes on the hydromt side ?

verseve:
Yes, this was only an issue on the Wflow side.

laurenebouaziz:
thanks @verseve !

After a git pull, I again tried the same run, unfortunately I get a different error (see below)

I guess it might be linked to an output variable of my toml file, but I don't see to which "field q" it refers to?

This is the toml file:
`p:\11205237-grade\wflow\wflow_meuse_julia\wflow_sbm_meuse_adapted_river_powlaw_subgrid\run_default\wflow_sbm_eobs_1980_2020.toml`

```
[ Info: Set `lateral.river.bankfull_elevation` using NetCDF variable `FloodplainZ`.
[ Info: Set `lateral.river.bankfull_depth` using NetCDF variable `RiverDepth`.
[ Info: Set `lateral.river.n` using NetCDF variable `N_River`.
[ Info: Parallel execution of kinematic wave, minimum stream order = `3`.
┌ Error: Wflow simulation failed
│   exception =
│    type ShallowWaterLand has no field q
│    Stacktrace:
│      [1] getproperty
│        @ .\Base.jl:33 [inlined]
│      [2] (::Base.BottomRF{typeof(getproperty)})(acc::Wflow.ShallowWaterLand{Float64}, x::Symbol)
│        @ Base .\reduce.jl:81
│      [3] afoldl(::Base.BottomRF{typeof(getproperty)}, ::NamedTuple{(:vertical, :lateral), Tuple{Wflow.SBM{Float64, 4, 5}, NamedTuple{(:subsurface, :land, :river), Tuple{Wflow.LateralSSF{Float64}, Wflow.ShallowWaterLand{Float64}, Wflow.ShallowWaterRiver{Float64, Wflow.SimpleReservoir{Float64}, Nothing}}}}}, ::Symbol, ::Symbol, ::Symbol)
│        @ Base .\operators.jl:535
│      [4] _foldl_impl
│        @ .\tuple.jl:263 [inlined]
│      [5] foldl_impl
│        @ .\reduce.jl:48 [inlined]
│      [6] mapfoldl_impl
│        @ .\reduce.jl:44 [inlined]
│      [7] #mapfoldl#214
│        @ .\reduce.jl:160 [inlined]
│      [8] #foldl#215
│        @ .\reduce.jl:178 [inlined]
│      [9] param(obj::NamedTuple{(:vertical, :lateral), Tuple{Wflow.SBM{Float64, 4, 5}, NamedTuple{(:subsurface, :land, :river), Tuple{Wflow.LateralSSF{Float64}, Wflow.ShallowWaterLand{Float64}, Wflow.ShallowWaterRiver{Float64, Wflow.SimpleReservoir{Float64}, Nothing}}}}}, fields::Tuple{Symbol, Symbol, Symbol})
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\io.jl:17
│     [10] out_map(ncnames_dict::Dict{Tuple{Symbol, Vararg{Symbol, N} where N}, String}, modelmap::NamedTuple{(:vertical, :lateral), Tuple{Wflow.SBM{Float64, 4, 5}, NamedTuple{(:subsurface, :land, :river), Tuple{Wflow.LateralSSF{Float64}, Wflow.ShallowWaterLand{Float64}, Wflow.ShallowWaterRiver{Float64, Wflow.SimpleReservoir{Float64}, Nothing}}}}})
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\io.jl:768
│     [11] prepare_writer(config::Wflow.Config, reader::Wflow.NCReader, modelmap::NamedTuple{(:vertical, :lateral), Tuple{Wflow.SBM{Float64, 4, 5}, NamedTuple{(:subsurface, :land, :river), Tuple{Wflow.LateralSSF{Float64}, Wflow.ShallowWaterLand{Float64}, Wflow.ShallowWaterRiver{Float64, Wflow.SimpleReservoir{Float64}, Nothing}}}}}, rev_inds::NamedTuple{(:land, :river, :reservoir, :lake), Tuple{Array{Int64, 2}, Array{Int64, 2}, Array{Int64, 2}, Nothing}}, x_nc::Array{Float64, 1}, y_nc::Array{Float64, 1}, nc_static::NCDatasets.NCDataset{Nothing}; maxlayers::Int64)
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\io.jl:834
│     [12] initialize_sbm_model(config::Wflow.Config)
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\sbm_model.jl:299
│     [13] run(config::Wflow.Config)
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:176
│     [14] (::Wflow.var"#178#179"{IOStream, Bool, Wflow.Config})()
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:155
│     [15] with_logstate(f::Function, logstate::Any)
│        @ Base.CoreLogging .\logging.jl:491
│     [16] with_logger(f::Function, logger::LoggingExtras.TeeLogger{Tuple{LoggingExtras.EarlyFilteredLogger{Logging.ConsoleLogger, Wflow.var"#172#175"}, LoggingExtras.EarlyFilteredLogger{LoggingExtras.MinLevelLogger{TerminalLoggers.TerminalLogger, Base.CoreLogging.LogLevel}, Wflow.var"#171#174"}}})
│        @ Base.CoreLogging .\logging.jl:603
│     [17] run(tomlpath::String; silent::Nothing)
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:151
│     [18] run
│        @ c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:144 [inlined]
│     [19] run()
│        @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:248
│     [20] top-level scope
│        @ none:1
│     [21] eval
│        @ .\boot.jl:360 [inlined]
│     [22] exec_options(opts::Base.JLOptions)
│        @ Base .\client.jl:261
│     [23] _start()
│        @ Base .\client.jl:485
└ @ Wflow c:\Users\bouaziz\.julia\dev\Wflow\src\Wflow.jl:162
```

verseve:
> I guess it might be linked to an output variable of my toml file, but I don't see to which "field q" it refers to?

It refers to `q` in your toml file related to overland flow. For the local inertial approach this should be `qx` and `qy`.

A TOML example:
```toml
[state.lateral.land]
h = "h_land"
h_av = "h_av_land"
qx = "qx_land"
qy = "qy_land"
volume = "vol_land"

[output.lateral.land]
h = "h_land"
qx = "qx_land"
qy = "qy_land"
````

**Comment by laurenebouaziz** [comment]
yes this is it! thanks @verseve it runs now :) looking forward to seeing the results after the weekend. 

btw, would you advice to start by applying the defaults for these params when running the model?
```
inertial_flow_alpha = 0.5        # alpha coefficient for model stability (default = 0.7)
froude_limit = true              # default is true, limit flow to subcritical-critical according to Froude number
h_thresh = 0.1                   # water depth [m] threshold for calculating flow between cells (default = 1e-03)

```

@DirkEilander, the config file which is now produced when updating the setup_river module includes the following lines:

```
[output.lateral.land]
q = "q_land"

[state.lateral.land]
q = "q_land"
```

so it will then be up to the user to change these line to qx and qy, if the user also adds the land_routing:
```
[model]
land_routing = "local-inertial"
```

File: hydromt_wflow/wflow.py [hunk 5]
```diff
@@ -235,95 +227,233 @@ def setup_basemaps(
...
+                smooth_len=smooth_len,
+                min_rivdph=min_rivdph,
+                min_rivwth=min_rivwth,
+                logger=self.logger,
+                **kwargs,
+            )
+            rmdict = {k: v for k, v in self._MAPS.items() if k in ds_riv1.data_vars}
+            self.set_staticmaps(ds_riv1.rename(rmdict))
+            # update config
+            # bankfull_depth for local inertial and h_bankfull for kin wave ?!
```
DirkEilander, line 377: @verseve Am I correct to understand that the river bankfull depth is referred to with different names in the toml file: For kinematic-wave `h_bankfull` and for local-inertial `bankfull_depth`. Would it make sense to make this naming consistent?

examples/wflow_build.ini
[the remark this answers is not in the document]
DirkEilander: - I have decided to keep the `wflow_dem` name as it seems to be used quite often in other scripts (e.g. the example notebooks) and might require a larger discussion on naming before changing.
- The `RiverZ` map has been deprecated in favor of `dem_subgrid` map which is derived when upscaling the flow directions and not hydrologically conditioned. 
- All the hydrological conditioning happens in `setup_hydrodem` where the user can select which elevation to apply the conditioning to. If based on `wflow_dem` the name becomes `hydrodem_avg`; if based on `dem_subgrid` > `hydrodem_subgrid`.
- I've added much explanation about the different option in the `setup_hydrodem` docstring

examples/wflow_piave_clip/wflow_sbm.toml
DirkEilander: See reply to earlier comments. We also set more wflow config options depending on the choice of routing.

hydromt_wflow/wflow.py
DirkEilander: Updated the docstring with much more info.

hydromt_wflow/workflows/river.py
DirkEilander: It's probably a width close to bank full that is in most datasets, but depending on landsat overpass etc.

hydromt_wflow/wflow.py
```diff
@@ -235,95 +235,245 @@ def setup_basemaps(
...
+            da_rivmsk=self.staticmaps[self._MAPS["rivmsk"]],
+            flwdir=self.flwdir,
+            connectivity=connectivity,
+            river_d8=True,
+            logger=self.logger,
+        ).rename(name)
+        self.set_staticmaps(ds_out)
+
+        # update toml model.river_routing
+        self.logger.debug(f'Update wflow config momdel.river_routing="{land_routing}"')
```
DirkEilander, line 457: @verseve Could you check if these settings L457-L475 are correct?

DirkEilander:
Not sure why the test are not triggered ..

hydromt_wflow/wflow.py
verseve: Yes, correct, in the toml file (and in the code, as part of the struct). I agree, would make sense to make this naming consistent!

hydromt_wflow/wflow.py
verseve: Yes, this looks correct, made some small corrections (typos) in https://github.com/Deltares/hydromt_wflow/pull/63/commits/11c95e3d4c3f4d96115483150f7c43fe29032315.

verseve (approved):
Nice work @DirkEilander!

hydromt_wflow/wflow.py
DirkEilander: thanks!

PR #67, opened 2021-12-15
WIP: Soilgrids

markhegnauer:
Working on reducing memory issues for soilgrids. Steps:
- Minimize number of times interpolate_na is called
- ...


laurenebouaziz:
hey Mark, the thing is that there can be missing values in the Soilgrids dataset and we first fill these values with the nearest neighbor interpolation. So if we remove this, we probably need to fill these values in another way. Now we might have some missing values in the resulting maps I guess?

verseve:
If I understand correctly, this is mainly about reducing memory usage. What about looking into the more direct causes, for example Dask and used Chunks?

markhegnauer:
@laurenebouaziz : Indeed, we don't want missing values for wflow. I thought to implement the interpolate_na function on the ds_out, just before writing (line 539), instead of on the input data with higher resolution. Just to test if this would result in a workable solution, since soilgrids is now blocking for our project because of memory issues. But unfortunately not working because of non-convergance error (SVD did not converge in Linear Least Squares).

@verseve : I can try now, but reason I started here was that I understood that the interpolate_na was identified as a memory consuming step in the process.

DirkEilander:
@markhegnauer Good initiative. To implement this we also need to update the soil maps in the test examples (ergo the failing test). It would be good to check how much it affects the soil maps for this example and perhaps also in some other cases. Any idea also on how it improves the memory usage?

@verseve Spatial operations such as nearest neighboring are very hard to implement "lazy" on spatial chunks, therefore the current approach is not lazy and indeed requires quite some memory. I agree with Mark that reducing the number of times we call the methods could be more 'low hanging fruit' compared to an lazy implementation. However if you have suggestions to improve the interpolate_na implementation that would be even better :)

verseve:
Hm, yes, I think the question is if the memory issue only occurs for the soilgrids part of the code with interpolate_na? If this is the case, then it is most probably caused by the vertical soil layer dimension. I think if the gaps are not too large, and chunks too small, nearest neighbor should be fine. The following stackoverflow question could be a starting point:

markhegnauer:
@DirkEilander the non-convergence error arrises in the do_linalg function in soilgrids. This happens when the data is missing, which makes sense to me. I'll try to capture this in the function (like you already capture the infinite value of log(y)). It indeed doesn't happen in the test dataset (that's why I commited already :)). And indeed, if the solution works, we need to update the test dataset, I saw the failing test indeed.

markhegnauer:
Just pushed an update. It's now working, in the sense that the maps can be generated for the Congo. I'll still check the differences, also for the testcase and will publish a new test dataset for the tests to pass. 

Code seems (much) faster up to the point where it goes into the do_linalg and especially the do_curve_fit functions, but I did not yet do a good test for it.

laurenebouaziz:
Nice Mark! It looks good (sorry I did not carefully read it before). 

Only thing I noticed is that in the sediment part, I think we should also add the two lines to interpolate ds_out for each variable at the end or decide to keep the interpolate before as it is only done for 3 maps, which are then used to compute the other required maps.

markhegnauer:
@laurenebouaziz thanks for the review. Just updated the sediment part as well.

DirkEilander:
Good work @markhegnauer. two small remark/suggestions:
- I noticed that the lake_b parameter for  one lake has changed in the test files, but that's not likely the result of this PR. Any idea why?
- Finally: could you add this change to the changelog before merging?

markhegnauer:
@DirkEilander about the lake_b parameter, no clue where that is coming from. Will have a look at it if I can figure out why. Regarding the changelog, will add and commit.

laurenebouaziz (approved):
All looks good now!

PR #70, opened 2021-12-23
Small things

hboisgon:
Small changes to allow local data sources in the different components, correct lake_b to avoid zero, add a fixed time units to allow wflow to run with different forcing files in one run.

### DIFF
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -769,9 +769,7 @@ class WflowModel(Model):
         }
 
         if lakes_fn not in self.data_catalog:
-            self.logger.warning(
-                f"Invalid source '{lakes_fn}', skipping setup_lakes."
-            )
+            self.logger.warning(f"Invalid source '{lakes_fn}', skipping setup_lakes.")
             return
         gdf_org, ds_lakes = self._setup_waterbodies(lakes_fn, "lake", min_area)
         if ds_lakes is not None:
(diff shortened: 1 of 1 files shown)


hydromt_wflow/wflow.py
```diff
@@ -768,7 +768,7 @@ def setup_lakes(self, lakes_fn="hydro_lakes", min_area=10.0):
             "input.lateral.river.lake.waterlevel": "LakeAvgLevel",
         }
 
-        if lakes_fn not in ["hydro_lakes"]:
+        if lakes_fn not in self.data_catalog:
```
DirkEilander, line 771: Should we throw a warning here? Later the `get_geodataframe` method will return an error if the dataset is not found which is more consistent with other methods.

DirkEilander: idem for glaciers in Line 1124

DirkEilander (changes requested):
@hboisgon Thanks for these improvements! "

It would be nice to directly address the following:
- update the docstring for setup_lakes and setup_glacies to explain in detail what data is expected and remove the statement that only hydro_lakes is supported
- generalize the requirements for both lakes and glaciers use uniform names for the variables (e..g now we use "simple_id for glaciers)

Lastly, it'd be great if you could add these changes to the changelog :)

hboisgon:
@DirkEilander : do you mind reviewing again this branch?

Some questions:

- I did not harmonize glacier simple_id. This would require an update of the data catalog. And that may mean that waterbody_id should also maybe change to a more global fid. To be honest I think this is linked to our more global hydromt issues of harmonizing names and variables units over hydromt and plugins. While filling in the docstrings I also realised units and names required by the different methods (lakes/reservoirs/gauges) are completely different. So prefer to have this done in another PR
- I updated the deprecation of gaugemap but not basin_shape. Are you sure basin_shape should be deprecated? It is already a combination of basin_map (if basin_name not in ds) and if it is it anyhow only just creates the shapes with apparently quite some 'treatment' lines around that would need to be repeated in setup_gauges then instead

Thanks in advance!

PR #75, opened 2022-02-09
prevent NaN FillValue in time coordinate of forcing 

### LINKED ISSUE #60: Wflow.jl and update forcing _FillValue
When update forcing in hydromt is used the time variable gets a `_FillValue = NaN` as attribute.  This gives problems when reading the file in Wflow.jl when dispatching on the type of the Vector containing time (`Vector{Union{Missing, Dates.DateTime}}`), which should be `Vector{Dates.DateTime}`.

My suggestion is to remove the `_FillValue` from the coordinates of the NetCDF file in hydromt.

This is probably default/standard when using `xarray`, however besides the issue with Wflow.jl it also violates cf standards.

DirkEilander:
- fixes #60

### DIFF
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1798,6 +1798,9 @@ class WflowModel(Model):
                 v: {"zlib": True, "dtype": "float32", "chunksizes": chunksizes}
                 for v in ds.data_vars.keys()
             }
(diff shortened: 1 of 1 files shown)


DirkEilander:
@hboisgon this is a very small PR fixing a specific single issue. Could you merge if you agree with the change?

hboisgon (approved):

PR #78, opened 2022-03-15
Switch to rioxarray

hboisgon:
Branch for when hydromt switches to rioxarray

Fix some dtype/nodata issue as rioxarray is stricter on fillvalue matching the dtype of the dataarray


hboisgon:
Tests are passing now because the rioxarray branch of hydromt is not merged yet. To be re-checked after the merging. (Locally RootingDepth is different and I am not sure why)

DirkEilander:
This branch is ready to merge (if the tests are successful). Even without the rioxarray implementation in the main this fixes:

- write **CF-compliant** staticmaps.nc and inmaps.nc
- correct float32 dtype for all landuse based maps (by changing values in all lookup tables to floats)
- uint8 dtype *wflow_rivers* and *wflow_streamorder* maps
- except for coordinates (incl *x_out* and *y_out*) all variables are saved with at most 32 bit depth
- new dtype and nodata arguments in **setup_constant_pars**

DirkEilander:
Can you do a final review @hboisgon ?

PR #92, opened 2022-07-19
Smooth rivlen

verseve:


hydromt_wflow/workflows/river.py
```diff
@@ -31,6 +31,8 @@ def river(
     river_upa=30.0,
     slope_len=2e3,
     min_rivlen_ratio=0.1,
+    min_rivlen=None,
```
DirkEilander, line 34: Should we not combine min_rivlen_ratio and min_rivlen?

hydromt_wflow/workflows/river.py
```diff
@@ -121,7 +128,23 @@ def river(
...
+    if min_rivlen != None:
+        logger.debug(f"Smooth river length (min length: {min_rivlen:.0f} m).")
+        flwdir_model = flw.flwdir_from_da(ds_model["flwdir"], mask=True)
+        rivlen_s = flwdir_model.smooth_rivlen(
+            rivlen=rivlen,
+            min_rivlen=min_rivlen,
+            smooth_cells=smooth_cells,
+            mask=riv_mask,
+        )
+        assert_equal = np.isclose(
```
DirkEilander, line 141: I think this is a good test, but wonder if this shouldn't be moved to the unit tests of pyflwdir instead?

DirkEilander:
alternative discussed: 
loop over cells smaller than min length
update lengths using average with iterative window size until min length is reached

hydromt_wflow/workflows/river.py
verseve: I agree, the test has been removed in commit https://github.com/Deltares/hydromt_wflow/pull/92/commits/298d83d2792e89bd975253b2d6f62d13afb50be5.

hydromt_wflow/workflows/river.py
verseve: Yes, as discussed we now use only `min_rivlen_ratio`.

DirkEilander:
[smooth rivlen PRs](https://trello.com/c/L9bjMkRj/17-smooth-rivlen-prs)

DirkEilander:
@verseve I've updated the tests and I think this is ready to merge. To apply the new smoothing we need the latest (unreleased) main branch of pylfwdir. I think it make sense to aim for a pyflwdir release after the sprint next week. Agreed?

DirkEilander:
I've just released pyflwidr v0.5.6 with the required smooth_rivlen method. This PR should bump the minimal pyflwdir version and than it's ready to go. see https://github.com/Deltares/pyflwdir/releases/tag/v0.5.6

@JoostBuitink: Can you bump the pyflwdir version in the pyproject.toml file and all environment *.yml files when you test the branch?
@hboisgon: It's not important for the course, but good to include it in the next release as it's been here for a while.

PR #100, opened 2022-08-02
update docs style and fix broken links

IMarth:
Update docs style due to issues after update of a package that is used for docs.


hboisgon (approved):
Thanks for the update @IMarth ! I did some minor changes in css to follow a recent one from Dirk in core and added an icon to Wflow website next to the GitHub one.

PR #101, opened 2022-08-03
no data value in setup_areamap to -999

laurenebouaziz:
I had an issue deriving a parameter which had values of 0 within the domain when running wflow. I would suggest to put no data to -999 instead of 0 in the function


hboisgon (commented):
Good catch!
With the switch to rioxarray, we are a bit more strict on dtypes. So if your col2raster is of dtype int8, -999 will not work as this would mean changing your data dtype. I think I would prefer 0 by default as it will be interpreted properly for most dtypes but add a new nodata argument to setup_areamap instead for users like you who do not want the 0 :)

PS: can you also update the changelog file with your new change ? I can also implement all this if you don't have time

Issue #97, opened 2022-08-02
check dates in setup_*_forcing


DirkEilander:
For the test Piave basin the dates of the forcing and toml file don't match.
Check where this error originates from and make sure that the toml file dates align with the forcing data in setup_*_forcing.

hboisgon:
I see indeed that when building the forcing for a larger period than available in the source_data no warning happens or the toml is also not updated.

The decision is to send a warning that not all requested dates are available and to update the start and end time in the config to the range available in the source_data

PR #102, opened 2022-08-04
Update times in setup_*forcing methods

hboisgon:
see #97

### DIFF
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1865,15 +1865,16 @@ class WflowModel(Model):
             for da in self.forcing.values():
                 if "time" in da.coords:
                     times = da.time.values
-                    if pd.to_datetime(start) < pd.to_datetime(times[0]):
+                    if start < pd.to_datetime(times[0]):
                         start = pd.to_datetime(times[0])
                         missings = True
-                    if pd.to_datetime(end) > pd.to_datetime(times[-1]):
+                    if end > pd.to_datetime(times[-1]):
                         end = pd.to_datetime(times[-1])
                         missings = True
-                        start = pd.to_datetime(times[0]).to_pydatetime()
-                        end = pd.to_datetime(times[-1]).to_pydatetime()
-            # Send warning and update config with new start and end time
+            # merge, process and write forcing
+            ds = xr.merge([da.reset_coords(drop=True) for da in self.forcing.values()])
+            ds.raster.set_crs(self.crs)
+            # Send warning, slice ds and update config with new start and end time
             if missings:
                 self.logger.warning(
                     f"Not all dates found in precip_fn changing starttime to {start} and endtime to {end} in the toml."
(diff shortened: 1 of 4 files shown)


DirkEilander (changes requested):
suggestion to do check on times in write_forcing

hydromt_wflow/wflow.py
```diff
@@ -1859,24 +1858,39 @@ def write_forcing(
                         self.write_config()  # re-write config
                         fn_out = fn_default_path
 
+            # Check if all dates between (starttime, endtime) are in all da forcing
+            start = pd.to_datetime(self.get_config("starttime"))
+            end = pd.to_datetime(self.get_config("endtime"))
+            missings = False
+            for da in self.forcing.values():
+                if "time" in da.coords:
```
DirkEilander, line 1866: time should always be in forcing right? Why did you add this if statement?

hydromt_wflow/wflow.py
```diff
@@ -1859,24 +1858,39 @@ def write_forcing(
...
+                    times = da.time.values
+                    if pd.to_datetime(start) < pd.to_datetime(times[0]):
```
DirkEilander, line 1868: start is already a datetime object, you can skip the `pd.to_datetime(start)` here. And same for end in next if-statement. See also comment on L1875

hydromt_wflow/wflow.py
```diff
@@ -1859,24 +1858,39 @@ def write_forcing(
...
+                        start = pd.to_datetime(times[0])
+                        missings = True
+                    if pd.to_datetime(end) > pd.to_datetime(times[-1]):
+                        end = pd.to_datetime(times[-1])
+                        missings = True
+                        start = pd.to_datetime(times[0]).to_pydatetime()
```
DirkEilander, line 1874: I think this line should be deleted as it is in the wrong if statement

hydromt_wflow/wflow.py
```diff
@@ -1859,24 +1858,39 @@ def write_forcing(
...
+                        end = pd.to_datetime(times[-1]).to_pydatetime()
```
DirkEilander, line 1875: I suggest to remove the `to_pydatatime()` here to always keep `end` (and `start`) as datetime object and only do the conversion when you update the config on L1881

hydromt_wflow/wflow.py
```diff
@@ -1859,24 +1858,39 @@ def write_forcing(
...
+                    f"Not all dates found in precip_fn changing starttime to {start} and endtime to {end} in the toml."
+                )
+                self.set_config("starttime", start.to_pydatetime())
+                self.set_config("endtime", end.to_pydatetime())
+                self.write_config()
+
             # merge, process and write forcing
             ds = xr.merge([da.reset_coords(drop=True) for da in self.forcing.values()])
+            ds.raster.set_crs(self.crs)
+            ds = ds.sel({"time": slice(start, end)})
```
DirkEilander, line 1888: I suggest to move this to the `if missings:` statement.

DirkEilander (changes requested):
Good improvement @hboisgon! 
One remark: I think L1874 should be removed / is misplaced. 
Otherwise some small suggestions.

hydromt_wflow/wflow.py
hboisgon: Apparently spatial_ref ends up as a da in self.forcing and it does not have time dimension

DirkEilander (approved):

Issue #109, opened 2022-10-04
Bug: pd.to_datetime on DatetimeNoLeap calendar


hboisgon:
New fix in #97 to check dates does not work on forcing data for which the datetime is of type cftime.DatetimeNoLeap
pd.to_datetime() method does not support that type:

>   File "C:\Users\boisgont\Miniconda3\envs\cst\lib\site-packages\hydromt_wflow\wflow.py", line 1871, in write_forcing
>     if start < pd.to_datetime(times[0]):
>   File "C:\Users\boisgont\Miniconda3\envs\cst\lib\site-packages\pandas\core\tools\datetimes.py", line 1078, in to_datetime
>     result = convert_listlike(np.array([arg]), format)[0]
>   File "C:\Users\boisgont\Miniconda3\envs\cst\lib\site-packages\pandas\core\tools\datetimes.py", line 402, in _convert_listlike_datetimes
>     result, tz_parsed = objects_to_datetime64ns(
>   File "C:\Users\boisgont\Miniconda3\envs\cst\lib\site-packages\pandas\core\arrays\datetimes.py", line 2224, in objects_to_datetime64ns
>     result, tz_parsed = tslib.array_to_datetime(
>   File "pandas\_libs\tslib.pyx", line 381, in pandas._libs.tslib.array_to_datetime
>   File "pandas\_libs\tslib.pyx", line 613, in pandas._libs.tslib.array_to_datetime
>   File "pandas\_libs\tslib.pyx", line 755, in pandas._libs.tslib._array_to_datetime_object
>   File "pandas\_libs\tslib.pyx", line 586, in pandas._libs.tslib.array_to_datetime
> TypeError: <class 'cftime._cftime.DatetimeNoLeap'> is not convertible to datetime

PR #110, opened 2022-10-05
Bug #109 forcing with DatetimeNoLeap

hboisgon:
See #109 : solution is to use xr.to_datetimeindex instead of pd.to_datetime to convert times in forcing DataArrays

### DIFF
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1867,7 +1867,10 @@ class WflowModel(Model):
             missings = False
             for da in self.forcing.values():
                 if "time" in da.coords:
-                    times = da.indexes["time"].to_datetimeindex().values
+                    if not isinstance(da.indexes["time"], pd.DatetimeIndex):
+                        times = da.indexes["time"].to_datetimeindex().values
+                    else:
+                        times = da.time.values
                     if start < pd.to_datetime(times[0]):
                         start = pd.to_datetime(times[0])
                         missings = True
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -763,9 +763,6 @@ class WflowModel(Model):
                     # try to get epsg number directly, important when writting back data_catalog
                     if hasattr(self.crs, "to_epsg"):
                         code = self.crs.to_epsg()
-                    elif hasattr(self.crs, "is_epsg_code"):
-                        if self.crs.is_epsg_code:
-                            code = int(self.crs["init"].lstrip("epsg:"))
                     else:
                         code = self.crs
                     kwargs.update(crs=code)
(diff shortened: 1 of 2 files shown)


hboisgon:
@DirkEilander : could you review this small bugfix? You reviewed this new change from #97 so you should already know more or less what it's about :) Many thanks!

hboisgon:
Tests should be fixed when merging https://github.com/Deltares/hydromt/pull/211 in hydromt

DirkEilander (approved):

hydromt_wflow/wflow.py
```diff
@@ -1867,7 +1867,10 @@ def write_forcing(
             missings = False
             for da in self.forcing.values():
                 if "time" in da.coords:
-                    times = da.time.values
+                    if not isinstance(da.indexes["time"], pd.DatetimeIndex):
```
DirkEilander, line 1870: Based on the xarray docs it seems that to_datetimeindex is a method of the `CFTimeIndex` object only. To avoid applying it to an time object type that does not have this method it would suggest to change it into:
`if isinstance(da["time"], xr.CFTimeIndex):`
or more general
`if hasattr(da['time'], 'to_datetimeindex'):`

DirkEilander (changes requested):

DirkEilander:
> Tests should be fixed when merging [Deltares/hydromt#211](https://github.com/Deltares/hydromt/pull/211) in hydromt

The test fails on another issue it seems, see https://github.com/Deltares/hydromt_wflow/issues/113

hydromt_wflow/wflow.py
```diff
@@ -761,8 +761,11 @@ def setup_gauges(
                 kwargs = {}
                 if isfile(gauges_fn):
                     # try to get epsg number directly, important when writting back data_catalog
-                    if self.crs.is_epsg_code:
-                        code = int(self.crs["init"].lstrip("epsg:"))
+                    if hasattr(self.crs, "to_epsg"):
+                        code = self.crs.to_epsg()
+                    elif hasattr(self.crs, "is_epsg_code"):
+                        if self.crs.is_epsg_code:
+                            code = int(self.crs["init"].lstrip("epsg:"))
```
DirkEilander, line 768: I think these lines can be removed because self.crs is always of pyproj.CRS type.

For the RasterModel see https://github.com/Deltares/hydromt/blob/075ac08d91f7e06bdc0504c0163d450e7b23a955/hydromt/models/model_grid.py#L181

For the Model see
https://github.com/Deltares/hydromt/blob/075ac08d91f7e06bdc0504c0163d450e7b23a955/hydromt/models/model_api.py#L1278

DirkEilander (approved):
Thanks for these fixes @hboisgon.
I think it's basically good to go. But strongly suggest to remove the lines that assume self.crs still has the rasterio CRS dtype as these will never be hit (see comment)

hydromt_wflow/wflow.py
hboisgon: This is true for hydromt >= 0.5.0 but not for older version I think. I don't mind removing I think we put pins everywhere already in the different env files.

PR #111, opened 2022-10-06
PR: Add mapping for ESA worldcover map

markhegnauer:
This pull request aims on providing HydroMT users with the option to also use ESA Worldcover maps for building their Wflow models.

The current parameters are basically 1-on-1 taken from VITO mapping table, so review is still needed. 

A simple test was done, but during that test I saw that the processing of the landuse map is much slower. This is logical, since the basemap has a 10x10 meter resolution, and thus is much higher resolution than GLOBCOVER, CORINE or VITO. We might expect some memory issues when going to (very) large domains. 

Next step could be deriving LU classes globally for 1x1 (and maybe also 2x2) km for ease of use.

image


markhegnauer:
I think we should also rethink the method of resampling for these high-resolution maps. The current (default) method for landuse resampling is "nearest". This gives, in my opinion, strange results, see figure. 

We could maybe switch the another resampling method for ESA WorldCover (or for all landcover maps), for example using the "mode" method. Seems to make it (even) slower though. A comparison between the 2 methods is shown in the figure. 

image

JoostBuitink:
Nice addition, thanks! While comparing values to the other mapping tables, I decided to add descriptions to the CORINE mapping table. 

Regarding the speed: we could also add an option to resample the LULC maps first to the model resolution, and then determine the parameters. Currently the parameters are determined at the original resolution of the LULC maps, and then resampled (by averaging) to the model resolution.

JoostBuitink (approved):

Issue #107, opened 2022-09-15
floodplain 1D schematization 


DirkEilander:
- floodplain volume as function of flood depth calculated per river branch at fixed intervals (e.g. 0.5 m) based on HAND with user inputs: depth intervals

based on:
- pyflwdir: a method that maps subgrid cells to river cell (at model resolution) outflow points based on D8, sorts the HAND values (in future also manning?) and returns the contributing area for fixed HAND intervals, see https://github.com/Deltares/pyflwdir/pull/29

Note:
- for future downscaling of flood depths to 2D maps it would be good to keep the highres HAND and mapping to branches

PR #123, opened 2022-11-09
add setup_river_floodplain method

DirkEilander:
see #107 

TODO:
- [x] create tests
- [x] add to docs
- [x] add toml setting in setup_river_floodplain
- [ ] add options for manning roughness 
- [x] merge setup_hydrodem & setup_river_floodplain?
- [x] release and update pyflwdir dependency.


DirkEilander:
Note that setting a list via the hydromt ini file does not work currently. A fix is proposed in https://github.com/Deltares/hydromt/pull/246

hydromt_wflow/workflows/river.py
```diff
@@ -291,6 +291,86 @@ def river_bathymetry(
...
+        drain=ds["uparea"].values >= river_upa, elevtn=ds["elevtn"].values
+    )
+
+    # calculate volume at user defined flood depths
+    # note that the ucat_map is not save currently but is required if
+    # we want to create a flood map by postprocessing
+    flood_depths = np.atleast_1d(flood_depths).ravel().astype(np.float32)  # force 1D
+    _, ucat_vol = flwdir.ucat_volume(idxs_out=idxs_out, hand=hand, depths=flood_depths)
+    # force minimum volume based on river width * length
+    min_vol = (
```
verseve, line 362: It might also be useful to check the `min_vol` when the previous segment has a trapezoidal form? 

Then the `min_vol` used for comparison should also increase. Note, that this check is already done in Wflow.jl, and for the Meuse and the Rhine models the volume for quite some segments is adapted. For example, for the Meuse a total of 20128 rectangular segments are adapted for all river cells.

DirkEilander: Thanks for this comment @verseve! I've addressed this in the latest commit here. Can you check this.

I've tested it by checking it with the following code. However i still get a warning message when running the latest wflow.jl of the 1d_floodplain branch, which I think is a rounding error, but not sure how to best go about it. I'll add a comment about this over at https://github.com/Deltares/Wflow.jl/pull/233/
```
import xarray as xr
import numpy as np
ds = xr.open_dataset(r'staticmaps.nc')
da_fldvol = ds['floodplain_volume'].where(ds['wflow_river']==1, 0)
# second diff should be >= zero
check_vol = (da_fldvol.diff('flood_depth', n=2) >= 0)
assert np.all(check_vol) 
```

verseve: I did check the changes in your latest commit, it looks good to me! I did add your suggested change for the warning message in Wflow.jl, I expect that with this change a warning message is not thrown. Is that the case?

DirkEilander: Thanks for the quick reply. I works as expected.

DirkEilander:
@JoostBuitink @laurenebouaziz @verseve 

There are a few last TODO's to finalize this PR. see description above. Regarding the question whether to merge `setup_hydrodem` and `setup_river_floodplain`, I'd like to get your opinion. Both methods deal with (different ways) to schematize the floodplains and to use the 1D floodplains you still need to run setup_hydrodem first. What is a clear way to divide the different processes? 

My suggestion:
- remove setup_hydrodem
- move the setup of 1D riverbed elevation (hydrodem) to `setup_rivers`. 
- move the setup of 2D elevation (i.e. updating the hydrodem) to setup_floodplains if the user chooses the 2D floodplain option.

Also, I cannot reproduce the failing test locally. Does it also fail for you? I'll try to do a clean checkout to see if there is a local file conflict somewhere..

laurenebouaziz:
@DirkEilander, yes I think this is a very good suggestion! Now we have:

[input.lateral.river]
bankfull_elevation = "hydrodem_subgrid_D4"

but in fact in case we have floodplain 1d routing, D4 or D8 does not matter because the river is D8 anyways. So we could have this standard as D8 if floodplain 1d is applied. 

If 1d2d is applied, it makes sense to update to D4 in a specific setup_2d_floodplains method (maybe good to have 2d included in the name of the method to make it very clear?)

@verseve, we were just discussing with @JoostBuitink that the toml also includes:
[input.vertical]
altitude = "wflow_dem"

how is this used in wflow and should this also be consistent (dem versus subgrid?)


About the test, I had updated the staticmaps and the local tests were working for me, but I will check again and add some more tests as we discussed!

hydromt_wflow/workflows/river.py
```diff
@@ -372,23 +372,27 @@ def river_floodplain_volume(
...
-        * ds_model["rivlen"].values[None, ...]
-        * flood_depths[:, None, None]
-    )
-    ucat_vol = (
-        np.where(riv_mask, np.maximum(ucat_vol, min_vol), -9999)
-        .round(0)
-    )
+    rivlen = np.maximum(ds_model["rivlen"].values, 1)  # avoid zero division errors
+    min_vol = ds_model["rivwth"].values * rivlen * flood_depths[0]
+    ucat_vol[0, :, :] = np.maximum(ucat_vol[0, :, :], min_vol).round(0)
```
verseve, line 377: I think it is probably better to leave out the rounding (`.round(0)`) in lines 377 and line 382. Without the rounding the provided flood volume error becomes a lot smaller. In Wflow.jl it is then possible to raise a warning based on volume (instead of width), which is more consistent (per river cell). I did test this and without the rounding it is fine to use of a volume threshold of 0.01 m3, without raising a warning. See also https://github.com/Deltares/Wflow.jl/pull/233/commits/f21fb7f9ecaa09e0d54a4a9843d3f2858a13f619.

verseve:
> @verseve, we were just discussing with @JoostBuitink that the toml also includes: [input.vertical] altitude = "wflow_dem"
> 
> how is this used in wflow and should this also be consistent (dem versus subgrid?)

This is not used by the `sbm` model type, it is only used by the `sbm_gwf` model type (as surface elevation).

verseve:
> but in fact in case we have floodplain 1d routing, D4 or D8 does not matter because the river is D8 anyways. So we could have this standard as D8 if floodplain 1d is applied.

Is it also then still possible to use the average dem value (`hydrodem_avg` ?) ?

hydromt_wflow/workflows/river.py
DirkEilander: OK, will change this!

DirkEilander:
> > but in fact in case we have floodplain 1d routing, D4 or D8 does not matter because the river is D8 anyways. So we could have this standard as D8 if floodplain 1d is applied.
> 
> Is it also then still possible to use the average dem value (`hydrodem_avg` ?) ?

Yes, we will keep this option. Or perhaps even remove the other option (subgrid outlet pixel elevation) as it is not a logical approach as we discussed this largely overestimates the floodplain volume, right? For the river bed elevation I will then also implement https://github.com/Deltares/hydromt_wflow/issues/126

laurenebouaziz:
> > > but in fact in case we have floodplain 1d routing, D4 or D8 does not matter because the river is D8 anyways. So we could have this standard as D8 if floodplain 1d is applied.
> > 
> > 
> > Is it also then still possible to use the average dem value (`hydrodem_avg` ?) ?
> 
> Yes, we will keep this option. Or perhaps even remove the other option (subgrid outlet pixel elevation) as it is not a logical approach as we discussed this largely overestimates the floodplain volume, right? For the river bed elevation I will then also implement #126

I think the subgrid option is good to keep, because it has a more realistic river slope profile right? with the average option, there was too much delay in the Meuse and Rhine models.

JoostBuitink:
add riv_uparea attribute to metadata of river map in staticmaps. Use this if no uparea is specified in the setup_floodplain function (raise error if different)

docs/changelog.rst
```diff
@@ -28,7 +31,7 @@ Fixed
 - fix error in **setup_reservoirs** when gdf contains no data in np.nanmax calculation for i.e. damheight #35
 - write_forcing with time cftime.DatetimeNoLeap #138 by removing slicing forcing if missings (not needed)
 - write_forcing automatic adjustment of starttime and endtime based on forcing content
-- When clipping a model from a model with multiple forcing files, a single netcdf is made in write_forcing and the * is removed from the filename. 
+- When clipping a model from a model with multiple forcing files, a single netcdf is made in write_forcing and the * is removed from the filename.
 
 Deprecated
 ----------
```
DirkEilander, line 37: Add that setup_hydrodem is removed

hydromt_wflow/wflow.py
```diff
@@ -392,96 +424,230 @@ def setup_rivers(
...
+    def setup_floodplains(
         self,
-        elevtn_map="wflow_dem",
-        river_routing="local-inertial",
-        land_routing="kinematic-wave",
+        hydrography_fn,
+        floodplain_type: str,
+        ### Options for 1D floodplains
+        river_upa: Optional[float] = None,
+        flood_depths: List = [0.5, 1.0, 1.5, 2.0, 2.5],
```
DirkEilander, line 462: is a max of 2.5m sufficient in most cases? I can image we need more layers.

DirkEilander (commented):
Really nice work @JoostBuitink 
I made two small suggestion, but in general I think this is good to go.
I'm however not allowed to approve it as I started the PR.

docs/changelog.rst
JoostBuitink: Thanks for spotting this, I have now added this

hydromt_wflow/wflow.py
JoostBuitink: Good point! I have added values up until 5.0m

JoostBuitink:
It looks like the glacier map was failing the tests. I reverted the glacier map back (in https://github.com/Deltares/hydromt_wflow/pull/123/commits/7f9909e150053382e47cb2dc58c377bfd88340cb) to the original map (see https://github.com/Deltares/hydromt_wflow/issues/143), as this now again represents expected behavior with `all_touched=True`

PR #127, opened 2022-11-16
Update example .ini files

JoostBuitink:


hboisgon (approved):
Thanks for the changes! Looks good :) I updated the changelog to document them

Issue #128, opened 2022-11-18
Possibility to add USLE parameters from own data


hboisgon:
Currently both parameters are directly computed from landuse or soil properties but a direct grid cannot be added.

Could use the opportunity for a generic setup_staticmaps_from raster or equivalent.
Ideally would use the future setup_grid_from_raster from the core when wflow moves to GridModel subclass.

PR #129, opened 2022-11-21
Added setup_staticmaps_from_raster 

hboisgon:
Can be used to solve #128 
This method allows to add wflow variables from direct raster data (eg USLE parameter or altitude, ksatver etc.)
The function also allows to update the toml file in order to still get ready to run model with the new staticmaps.


JoostBuitink (approved):
Very nice addition, and great to see some tests added as well!

PR #130, opened 2022-11-21
Introducing Global Water Watch functionality for reservoir parameter calculation

jurianbeunk:
Hi @hboisgon, 

I've made updates on the code, added tests & documentation. Now ready for **review**. 

As you'll see in the changelog, I've:
1. changed the settings in [setup_reservoirs]: removed priorityjrc and usehe options, replaced with timeseries_fn.
2. made sure all downloaded time series are saved to root and indexed on Hylak ID.
3. added a test checking if all reservoirs parameters are present and contain non-null values.
4. added information to the documentation.

All tests ran successfully on my computer, including the newly developed one. 

Had fun working on this. Let me know if you agree with my decisions or if you have any questions!

Cheers, 
Jurian


jurianbeunk:
@hboisgon I've made a first draft of the scripts. Happy to discuss with you (as might known I normally start work around 8:00 so feel free to ring me up if you have time)!

hboisgon (approved):
Thanks a lot @jurianbeunk for the addition of gww! Looks really good :) I only updated that we use timeseries_fn = None by default in the code (so case that your reservoir_fn actually has all the parameters you need) and updated the templates build files to use gww.
Thanks again for a nice contribution!

PR #133, opened 2022-11-23
update install guide

DirkEilander:
- remove hydromt-wflow dependencies from environment.yml
- make installing using this file the recommended way
- update readme for examples


JoostBuitink (approved):
Nice corrections and improvements!

Issue #138, opened 2022-12-19
Error when creating forcing using datetime noleap type


Jaapel:
Hi all!

While preparing a forcing from a datasource that has a noleap type, when using pandas>1.4 (1.5.2 in my case), I encounter the following error:

```
Traceback (most recent call last):
  File "/root/work/.snakemake/scripts/tmp9je2e4p2.downscale_climate_forcing.py", line 97, in <module>
    mod.update(opt=update_options)
  File "/opt/conda/lib/python3.10/site-packages/hydromt/models/model_api.py", line 291, in update
    self._run_log_method(method, **opt[method])
  File "/opt/conda/lib/python3.10/site-packages/hydromt/models/model_api.py", line 153, in _run_log_method
    return func(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/hydromt_wflow/wflow.py", line 1982, in write_forcing
    ds = ds.sel({"time": slice(start, end)})
  File "/opt/conda/lib/python3.10/site-packages/xarray/core/dataset.py", line 2565, in sel
    query_results = map_index_queries(
  File "/opt/conda/lib/python3.10/site-packages/xarray/core/indexing.py", line 183, in map_index_queries
    results.append(index.sel(labels, **options))
  File "/opt/conda/lib/python3.10/site-packages/xarray/core/indexes.py", line 442, in sel
    indexer = _query_slice(self.index, label, coord_name, method, tolerance)
  File "/opt/conda/lib/python3.10/site-packages/xarray/core/indexes.py", line 207, in _query_slice
    indexer = index.slice_indexer(
  File "/opt/conda/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6602, in slice_indexer
    start_slice, end_slice = self.slice_locs(start, end, step=step)
  File "/opt/conda/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6810, in slice_locs
    start_slice = self.get_slice_bound(start, "left")
  File "/opt/conda/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6726, in get_slice_bound
    return self._searchsorted_monotonic(label, side)
  File "/opt/conda/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6672, in _searchsorted_monotonic
    return self.searchsorted(label, side=side)
  File "/opt/conda/lib/python3.10/site-packages/pandas/core/base.py", line 1298, in searchsorted
[...]

PR #142, opened 2023-01-10
Bugfix #138 in write_forcing

hboisgon:
Correct bug with forcing with cftime.DatetimeNoLeap

Problem was coming that if forcing timespan was shorter than intented wflow starttime and endtime (old missings flag), we would sel the forcing dataset:
https://github.com/Deltares/hydromt_wflow/blob/04fad113fa9a8763c380bb432b497e0bcb837b31/hydromt_wflow/wflow.py#L1982

The bug was here that start and end are pd.datetime and the slice does not work if ds.time is of type cftime.datetime. It can be fixed by converting start and end to strings first with start.strftime(format).
I don't really think we need the slicing though so I deleted it.

What I also found is that the starttime and endtime can have a different timestamp than the forcing (eg 2010-01-01 08:00:00 instead of 2010-01-01 00:00:00) so I added an extra check to correct the starttime and endtime to match better what is inside the forcing file


hboisgon:
@Jaapel can you check that this solves your issue ?

Jaapel:
I do not get the datetime error anymore 👍

hboisgon:
Would be nice to test hydromt_wflow (and hydromt) with cftime data maybe when cmip6 from gcs is implemented.
While digging I found than in xarray.open_dataset there is a use_cftime flag which is None by default. This means that times will be decoded to np.datetime64[ns] where possible and else to cftime.datetime:
https://docs.xarray.dev/en/stable/generated/xarray.open_dataset.html

As these two objects are quite different in functionalities I wonder if we maybe should not be more strict in hydromt and maybe always use cftime? @DirkEilander what do you think?

DirkEilander:
What are the different methods of the numpy datetime and cftime objects? I'm not sure if forcing all to be cftime will help or create more problems. Perhaps good to raise an issue in the core and investigate the benefits and limitations of both. If can easily accommodate for both that would be my preferred option.

hboisgon:
I agree it's not optimal to change but still maybe worth a discussion in HydroMT so I'll open one there. In the meantime, could you review this PR?

hydromt_wflow/wflow.py
```diff
@@ -1964,22 +1965,27 @@ def write_forcing(
                         times = da.time.values
                     if start < pd.to_datetime(times[0]):
                         start = pd.to_datetime(times[0])
-                        missings = True
+                        correct_times = True
+                    elif start not in times:
+                        start = pd.to_datetime(times[times >= start][0])
```
laurenebouaziz, line 1970: I think if starttime not in forcing_fn, the code here should be:
start = pd.to_datetime(times[0])
to make the starttime similar to the starttime of forcing_fn

times[times >= start][0]
the above line would result in an error as start is not in times. 

I will commit this change

laurenebouaziz:
the workflow does not work for 360 days calendars.
when `da.indexes['time'].calendar != "360_day"` the `.to_datetimeindex() ` does not work because of the presence of 29th of february each year: 
`*** ValueError: Cannot convert date 1970-02-29 00:00:00 to a date in the standard calendar.  Reason: day is out of range for month.`

Maybe good to further discuss how we want to deal with this.

hydromt_wflow/wflow.py
laurenebouaziz: actually the same issue occurred if end time was lower than times[-1] (so also not in times), so unless I overlooked somehting, I think the code can be simplified to:

```
                    if (start < pd.to_datetime(times[0])) | (start not in times):
                        start = pd.to_datetime(times[0])
                        correct_times = True
                    if (end > pd.to_datetime(times[-1])) | (end not in times):
                        end = pd.to_datetime(times[-1])
                        correct_times = True
```

this is done in commit 6c4f1317e207e04355ad4ed3ce6c763ae095a6c1

laurenebouaziz:
discussed with @hboisgon that we only correct dates in toml for standard calendars. see 0b3f977326867c831c2ba67490a10547ebad3916

PR #145, opened 2023-01-10
fix #144

### LINKED ISSUE #144: bug with wrong nodata value in workflows/basemaps.py#L150
This line is only used when the model region has no (sub)basin outlets. For instance when deriving a model based on a bbox / interbasin.
It can be reproduced with the following command line

`hydromt build wflow ./wflow_grande_daily_05_30sec_interb "{'interbasin': [-63.4, -18.3, -62.7, -17.4], 'strord': 4}" --dd -vv`

DirkEilander:


hboisgon (approved):
Looks good! 
Maybe update the changelog? This is a small issue so maybe not necessary

DirkEilander:
Add the fix to the changelog. Merging and closing the issue now.

Issue #146, opened 2023-01-10
bug with river workflow in the setup_rivers


couasnonanais:
Saving of rivwth and qbankfull in the river function is incorrectly set from the riv_mask not being saved as a boolean.

hboisgon:
Closed with merging #147

PR #147, opened 2023-01-10
Bug fix of river.py. Ref #146

couasnonanais:


DirkEilander (approved):
Thanks for investigating and fixing this issue @couasnonanais !

PR #152, opened 2023-02-09
Updated hydromt CLI calls using region and res arguments

Tjalling-dejong:
I updated all the HydroMT CLI calls that use region and resolution arguments in the examples. As well as adding res arguments to wflow_build.ini and wflow_sediment_build.ini

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,9 +25,9 @@ jobs:
         uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: "3.9"
-          mamba-version: "*"
+          miniforge-variant: Mambaforge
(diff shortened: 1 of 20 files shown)


DirkEilander:
Looks good @Tjalling-dejong 

I additionally 
- fixed the test environment 
- fixed the test that call build directly
- updated the env yml files and pyproject.toml to reflect changes in core (e.g. remove pygeos)

This can be merged as soon as hydromt core is released.

The remaining error has to do with the wflow_glaciermap ...

DirkEilander (approved):
Thanks @Tjalling-dejong for the changes.
In addition I have also updated the user guide (docs) and tests.

PR #158, opened 2023-03-09
Rating curve for lakes

hboisgon:
support rating curve tables for lake modelling within hydromt wflow
Allow to prepare lake modelling using rating curves when available:

- [x] rating curves supplied in simple csv format with lake ID in the filename : need to adapt GeoDataset reader to read multiple csv files or add an {index} key in the path similar to {year/month} to keep using DataFrame 
- [x] add to a hydromt object: not sure which one yet? I think this may actually be lacking from hydromt? So I created a self.tables new object which is a dict of DataFrames
- [x] allow for direct use of lake parameters from lakes_fn eg for LakeStoreFunc and LakeOutflowFunc
- [x] add readers and writers (should allow updating automatically)
- [x] update clipping of lake reservoirs to potentially remove some of the tables


hboisgon:
Hi @DirkEilander 
I tried to add support of lake rating and storage curves in setup_lakes.

Before a real review I came across two main challenges that I would like to discuss:

- **Reading several csv files**: I think more often than not rating curves will be stored in several csv files rather than the one with column names containing IDs for level, volume and discharge. I am not fully sure if GeoDataset fully make sense either and if it does you would still need to allow reading from several csv files. Instead I came up with a {index} known key in the data catalog path similar to [variable, month, year]. Would love to get your thoughts on that?
- **Storing tabular data in Model**: I really tried to think where to store the lake rating curve table but config/forcing/staticmaps did not make sense. I think we may be lacking a tables object in HydroMT which would be a dictionnary of pandas.DataFrame. What is your thought on this? 

Thanks in advance for your thoughts!

DirkEilander:
A quick reply:

> * **Reading several csv files**: I think more often than not rating curves will be stored in several csv files rather than the one with column names containing IDs for level, volume and discharge. I am not fully sure if GeoDataset fully make sense either and if it does you would still need to allow reading from several csv files. Instead I came up with a {index} known key in the data catalog path similar to [variable, month, year]. Would love to get your thoughts on that?

I can image we do support this in GeoDataset specifically with a new driver. My first idea would be to make it similar to [gdaltindex](https://gdal.org/programs/gdaltindex.html) where each geometry has a 'tileindex' column with a reference to a raster tile. Here we could support a similar 'tableindex' column with a reference to a table. Your fix with {index} would also be a good alternative, but we need to keep the complexity of the data catalog in mind.  

> * **Storing tabular data in Model**: I really tried to think where to store the lake rating curve table but config/forcing/staticmaps did not make sense. I think we may be lacking a tables object in HydroMT which would be a dictionnary of pandas.DataFrame. What is your thought on this?

I think we can solve this model specific, i.e. make a new property for wflow if that's needed. Not sure yet there is a global need for this.

hboisgon:
Well in the end in order to not use a new {index} known key, then I used a placeholder and the user then needs to list all the rating tables files / catalog entries to use. So if 20 lake rating curve files, then 20 entries for rating_curve_fns argument. 
I think a little less nice on the user side but compared to * this means that at least the data entry can be read on its own.

I think the optimal solution as the DataFrame type returns a single dataframe and GeoDataset does not support yet to read from several csv files.

hydromt_wflow/wflow.py
```diff
@@ -1103,18 +1118,81 @@ def setup_lakes(self, lakes_fn="hydro_lakes", min_area=10.0):
...
+        rating_dict = dict()
+        if rating_curve_fns is not None:
+            rating_curve_fns = np.atleast_1d(rating_curve_fns)
+            # assume lake index will be in the path
+            # Assume one rating curve per lake index
+            for id in gdf_org["waterbody_id"].values:
+                # Find if id is is one of the paths in rating_curve_fns
+                if any([str(id) in fn for fn in rating_curve_fns]):
+                    # Update path based on current waterbody_id
+                    rating_fn = [fn for fn in rating_curve_fns if str(id) in fn][0]
```
DirkEilander, line 1163: The check with "id in fn" is prone to errors as id 1 will also appear in at a file with index 101.
I suggest to parse the IDs from the names ones and then look for exact values.
Perhaps requiring a specific nomenclature (e.g. _{index}.csv) would make it more robust.

hydromt_wflow/workflows/waterbodies.py
```diff
@@ -482,3 +482,158 @@ def reservoirattrs(gdf, timeseries_fn=None, perc_norm=50, perc_min=20, logger=lo
...
+                    gdf.loc[gdf["waterbody_id"] == id, "LakeOutflowFunc"] = 1
+                    df_rate = df_rate[["elevtn", "discharge"]].dropna(
+                        subset=["elevtn", "discharge"]
+                    )
+                    df_rate.rename(
+                        columns={"elevtn": "H", "discharge": "Q"},
+                        inplace=True,
+                    )
+                    # Repeat Q for the 365 JDOY
+                    df_q = pd.concat([df_rate.Q] * (366), axis=1, ignore_index=True)
```
DirkEilander, line 602: Do we need a table with a column for each day of the year if these are all the same?
And what happens with in leap years?

DirkEilander (changes requested):
Nice work @hboisgon.
I've verified the code and tests work, but not tested it with wflow.
There is one step with connecting the IDs of the files and lakes that I think needs to be fixed, otherwise it looks ok.

hydromt_wflow/workflows/waterbodies.py
hboisgon: Yes this is the Wflow format... Also I think they then take care of the leap years (these are like cyclic tables for JDOY). I also tested with wflow.

hydromt_wflow/wflow.py
```diff
@@ -1154,13 +1156,18 @@ def setup_lakes(
...
             for id in gdf_org["waterbody_id"].values:
                 # Find if id is is one of the paths in rating_curve_fns
-                if any([str(id) in fn for fn in rating_curve_fns]):
+                if id in fns_ids:
                     # Update path based on current waterbody_id
-                    rating_fn = [fn for fn in rating_curve_fns if str(id) in fn][0]
+                    for i, idx in enumerate(fns_ids):
+                        if idx == id:
+                            rating_fn = rating_curve_fns[i]
+                            break
```
DirkEilander: you can use the index funtion of lists to do this `i = fn_ids.index(id)`

hydromt_wflow/wflow.py
```diff
@@ -1103,18 +1119,88 @@ def setup_lakes(self, lakes_fn="hydro_lakes", min_area=10.0):
...
+            return
+        rmdict = {k: v for k, v in self._MAPS.items() if k in ds_lakes.data_vars}
+        ds_lakes = ds_lakes.rename(rmdict)
+
+        # If rating_curve_fn prepare rating curve dict
+        rating_dict = dict()
+        if rating_curve_fns is not None:
+            rating_curve_fns = np.atleast_1d(rating_curve_fns)
+            # Find ids in rating_curve_fns
+            fns_ids = [int(fn.split("_")[-1].split(".")[0]) for fn in rating_curve_fns]
```
DirkEilander, line 1160: Would a try-except statement with a meaningfull error message help here in case the ids cannot be parsed as integers from the filenames

DirkEilander (approved):
Thanks for the update @hboisgon. I added two minor suggestions, but generally it's good to go.

PR #161, opened 2023-03-15
Support local DEM

hboisgon:
Support local DEM data and add notebook on examples how to create flwdir/uparea/strord based on elevation.


hboisgon:
@DirkEilander : this PR/issues were on you but I needed them earlier than what I thought for a project :)
The basin index issue was indeed a problem so this is fixed and I removed the error is crs was not 4326.
All the rest seem to work fine!

So still to do is:

- Notebook to prepare from elevation flwdir/uparea/basins/strord
- Docs / changelog / test?

hboisgon:
@DirkEilander I simplified the example notebook so I think it's ready for review!

DirkEilander (approved):

Issue #94, opened 2022-07-27
Move parameter mapping tables and add yml file


JoostBuitink:
Move the parameter mapping tables from inside the code to the main directory of hydromt_wflow in order to improve transparency. Additionally, provide a yml file to read those files using the new DataFrameAdapter in hydromt. 

The argument containing the location/name of the mapping table will become a required argument in the future. For now, when no name is given, we'll revert back to the original way (linking name of lulc maps to mapping table), and throw a DeprecationWarning to prepare users for this breaking change.

hboisgon:
I had a go at this in the hydromt_delft3dfm plugin. Examples are:

- data catalog: https://github.com/Deltares/HYDROLIB/blob/hydromt/hydrolib/hydromt_delft3dfm/data/parameters_data.yml
- inclusion in the code: https://github.com/Deltares/HYDROLIB/blob/hydromt/hydrolib/hydromt_delft3dfm/dflowfm.py

https://github.com/Deltares/HYDROLIB/blob/d768984b8bd6f3da20757fa7300accbaa805265c/hydrolib/hydromt_delft3dfm/dflowfm.py#L116-L118

https://github.com/Deltares/HYDROLIB/blob/d768984b8bd6f3da20757fa7300accbaa805265c/hydrolib/hydromt_delft3dfm/dflowfm.py#L168

PR #168, opened 2023-03-23
Remove hardcoded mapping tables and add those to a yml file

JoostBuitink:
This removes the hardcoded links to .csv files and moves those files to a data_catalog.yml file. The files remain in the data folder, and that is also where the .yml file is located. This file is automatically added to the WflowModel when this class is created.

See https://github.com/Deltares/hydromt_wflow/issues/94


hydromt_wflow/wflow.py
```diff
@@ -622,21 +625,21 @@ def setup_lulcmaps(
...
         self.logger.info(f"Preparing LULC parameter maps.")
         if lulc_mapping_fn is None:
-            fn_map = join(DATADIR, "lulc", f"{lulc_fn}_mapping.csv")
+            fn_map = f"{lulc_fn}_mapping"
         else:
             fn_map = lulc_mapping_fn
-        if not isfile(fn_map):
-            self.logger.error(f"LULC mapping file not found: {fn_map}")
-            return
+        if not isfile(fn_map) and fn_map not in self.data_catalog:
```
hboisgon, line 631: To avoid conflict with user data (eg user defined corine_mapping source), maybe we should name the parameters data corine_mapping_default or something so here if fn_map is not found try for {lulc_fn}_mapping_default as well before returning an error?
Because the parameters_data catalog is read last  think it has priority on source names compared to the other data catalogs... When in core we solve issue [#329](https://github.com/Deltares/hydromt/issues/329) then we can be more precise.

hboisgon (commented):
Sorry it took so long to review. Just one comment on the names in the default parameters_data.yml. I'm affraid source names might conflict so I added a suggestion to try and limit the cases where this can occur

hboisgon (approved):

PR #184, opened 2023-06-09
Rootzone clim

laurenebouaziz:
implement an alternative way to calculate the rootingdepth for wflow_sbm based on hydrometeorological data (for historical and future conditions), which follows the workflow described in Bouaziz et al. 2022.

Bouaziz, L. J. E., Aalbers, E. E., Weerts, A. H., Hegnauer, M., Buiteveld, H., Lammersen, R., Stam, J., Sprokkereef, E., Savenije, H. H. G., and Hrachowitz, M.: Ecosystem adaptation to climate change: the sensitivity of hydrological predictions to time-dynamic model parameters, Hydrol. Earth Syst. Sci., 26, 1295–1318, https://doi.org/10.5194/hess-26-1295-2022, 2022.


hydromt_wflow/workflows/rootzoneclim.py
```diff
@@ -0,0 +1,857 @@
...
+                 start_hydro_year,
+                 start_field_capacity,
+                 LAI,
+                 rooting_depth,
+                 correct_cc_deficit,
+                 chunksize,
+                 missing_days_threshold,
+                 logger=logger):
+    """
+    Returns root zone storage parameter for current observed and (optionally
```
hboisgon, line 527: Could you maybe then base this on the current wflow forcing file ? Or external data? Does anything happen if you have the three forcing available or do you just add a layer per forcing? If the later the user could also just call the function three times. This would reduce complexity

hboisgon (changes requested):
Looks like quite some work! Nice! I have a few comments especially on docstrings and what is the outcome of this function (is it actual wflow sbm input map? or something else?).
I didn't really review the workflow function very deeply yet, I can if needed.

Nice that tests are here! But the changelog and documentation (API and wflow model components) are missing.

hydromt_wflow/wflow.py
```diff
@@ -1584,6 +1584,158 @@ def setup_temp_pet_forcing(
...
+        #make sure dsrun overlaps with ds_obs, otherwise give error
+        if dsrun.time[0] < ds_obs.time[0]:
+            dsrun = dsrun.sel(time=slice(ds_obs.time[0], None))
+        if dsrun.time[-1] > ds_obs.time[-1]:
+            dsrun = dsrun.sel(time=slice(None, ds_obs.time[-1]))
+        if len(dsrun.time) == 0:
+            self.logger.error("No overlapping period between the meteo and observed streamflow data")
+
+        # Run the rootzone clim workflow
+        dsout, gdf = workflows.rootzoneclim(
```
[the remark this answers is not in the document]
laurenebouaziz: yes, I added a logger.error message, is this ok?

hydromt_wflow/workflows/rootzoneclim.py
laurenebouaziz: not sure I understand what you mean here. If you want to calculate the rootzone storage capacity for the future, you need to use results of ds_obs and potentially ds_cc_ref, so it wouldn't be possible to call the function 3 times with a different ds_obs. otherwise, it would indeed decrease complexity...

hydromt_wflow/wflow.py
```diff
@@ -1584,6 +1584,158 @@ def setup_temp_pet_forcing(
...
+            Gridded timeseries with the observed forcing. The default is "inmaps".
+            Expects to have variables "precip" and "pet".
+        forcing_cc_hist_fn : str, optional
+            Gridded timeseries with the simulated historical forcing, based on a climate
+            model. Expects to have variables "precip" and "pet". The default is None.
+        forcing_cc_fut_fn : str, optional
+            Gridded timeseries with the simulated climate forcing, based on a
+            climate model. Expects to have variables "precip" and "pet". 
+            The default is None.
+        chunksize : int, optional
```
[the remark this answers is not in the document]
laurenebouaziz: actually the chunksize here is not used to read in the data, but rather in one of the calculation steps in rootzoneclim.py (line 751 after commit 1ddf73bcafc1b6e8a03b0ca1d1f5b8cb8d1327de), once the averages of precip and pet per subcatchment are being calculated.

laurenebouaziz:
thanks a lot @hboisgon for your review! I made the changes (mostly more details and corrections in the docstrings and expansion of the test), could you have another look?
I also updated the changelog and api.rst, do more documentation files need to be updated?

hboisgon:
> thanks a lot @hboisgon for your review! I made the changes (mostly more details and corrections in the docstrings and expansion of the test), could you have another look? I also updated the changelog and api.rst, do more documentation files need to be updated?

Thanks! And yes also the table in docs/user_guide/wflow_model_setup.rst (which apparently is currently broken on the docs page...)

hydromt_wflow/wflow.py
```diff
@@ -1584,6 +1585,189 @@ def setup_temp_pet_forcing(
...
+        start_field_capacity : str, optional
+            The end of the wet season / commencement of dry season. This is the
+            moment when the soil is at field capacity, i.e. there is no storage
+            deficit yet. The default is 'Apr'. 
+        LAI : bool, optional
+            Determine whether the LAI will be used to determine Imax. The
+            default is False.
+            If set to True, requires to have run setup_laimaps. 
+        rooting_depth : bool, optional
+            Determines whether the rooting depth (rootzone storage /
```
hboisgon, line 1685: In the docstrings you say this method sets the RootingDepth but here you  provide an optional True/False flag for it?

hydromt_wflow/wflow.py
```diff
@@ -1584,6 +1585,189 @@ def setup_temp_pet_forcing(
...
+        The main assumption is that vegetation adapts its rootzone storage capacity 
+        to overcome dry spells with a certain return period (typically 20 years for forest ecosystems). 
+        In response to a changing climtate, 
+        it is likely that vegetation also adapts its rootzone storage capacity, 
+        thereby changing model parameters for future conditions.
+        This method also allows to estimate the change in rootzone storage capacity 
+        in response to a changing climate. 
+
+        As the method requires precipitation and potential evaporation timeseries, it may be useful to run this method as an update step in the setting-up of the hydrological model, once the forcing files have already been derived. 
+        In addition the setup_soilmaps method is also required if RootingDepth are to be calculated.
```
hboisgon, line 1629: and setup_laimaps?

hydromt_wflow/wflow.py
```diff
@@ -1584,6 +1585,189 @@ def setup_temp_pet_forcing(
...
+        Bouaziz, L. J. E., Aalbers, E. E., Weerts, A. H., Hegnauer, M., Buiteveld, 
+        H., Lammersen, R., Stam, J., Sprokkereef, E., Savenije, H. H. G. and 
+        Hrachowitz, M. (2022). Ecosystem adaptation to climate change: the 
+        sensitivity of hydrological predictions to time-dynamic model parameters, 
+        Hydrology and Earth System Sciences, 26(5), 1295-1318. DOI: 
+        10.5194/hess-26-1295-2022.
+
+        Adds model layer:
+
+        * **rootzone_storage_{forcing}_{RP}** map: rootzone storage capacity [mm of water] estimated from hydroclimatic data {forcing: obs, cc_hist or cc_fut} for different return periods RP.
```
hboisgon, line 1642: These maps, are they ever used by wflow_sbm? Else I'm starting to wonder if we should not add them to auxmaps and starts using that object in hydromt_wflow? I have that as well for the different Ksat per layer in staticmaps too. In hydromt we produce a lot of extra maps that are actually not used and I think confuse users when we give them an example wflow model. What do you think?

Else it's something I can look into maybe in the GridModel PR where we also switch from staticmaps to grid.

hydromt_wflow/wflow.py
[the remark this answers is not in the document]
hboisgon: Perfect!

hydromt_wflow/workflows/rootzoneclim.py
```diff
@@ -0,0 +1,875 @@
...
+    Returns
+    -------
+    ds_sub : xr.Dataset
+        Same as above, but containing the storage deficits per time step for all
+        forcing types.
+    """         
+    #make sure the order of the coordinates is always the same. 
+    # Calculate it per forcing type
+
+    ds_sub["storage_deficit"] = (
```
hboisgon, line 247: You could think of using hydromt.raster.full_like method instead (also in other places like this). Advantage is that you can choose to keep the new array lazy and np.nan can be directly set as the data values

hydromt_wflow/workflows/rootzoneclim.py
```diff
@@ -0,0 +1,875 @@
...
+        ("index", "forcing_type", "time"),
+        np.zeros(
+            (len(ds_sub["index"]), len(ds_sub["forcing_type"]), len(ds_sub["time"]))
+            )*np.nan
+        )
+
+    for forcing_type in ds_sub["forcing_type"].values:
```
hboisgon, line 254: In this whole loop you are switching to numpy rather than xarray. Is there a specific reason? Because here you are still working with daily gridded data correct? So this will be really hard on memory

hydromt_wflow/workflows/rootzoneclim.py
```diff
@@ -0,0 +1,875 @@
...
+
+    The method is based on the estimation of maximum annual storage deficits based on precipitation and estimated actual evaporation time series, 
+    which in turn are estimated from observed streamflow data and long-term precipitation and potential evap. data, as explained in Bouaziz et al. (2022).
+    The main assumption is that vegetation adapts its rootzone storage capacity to overcome dry spells with a certain return period (typically 20 years for forest ecosystems). 
+    In response to a changing climtate, it is likely that vegetation also adapts its rootzone storage capacity, thereby changing model parameters for future conditions.
+    This method also allows to estimate the change in rootzone storage capacity in response to a changing climate. 
+    
+    Parameters
+    ----------
+    dsrun : xr.Dataset
```
hboisgon, line 550: I see in check_inputs the variable should be named run ? We don't really have good hydromt conventions there yet which is why I opened an issue but maybe discharge (m3/s) might be more intuitive? I wonder if it's not what we use with chelsea

hboisgon (commented):
Thanks for the additions @laurenebouaziz ! Much more clear with the added docs. I added a few more comments/suggestions

hydromt_wflow/workflows/rootzoneclim.py
laurenebouaziz: Actually I had chosen the name run, based on the data conventions stated under: 
https://deltares.github.io/hydromt/latest/user_guide/data_conventions.html# (in hydrologgy section -- run for surface water runoff (overland flow + river discharge) in m3/s). Let me know if you would rather change it?

hydromt_wflow/workflows/rootzoneclim.py
laurenebouaziz: the thing was that when working directly with xarray the code was a lot lot slower, this is why we decided with @RubenImhoff to switch to numpy

laurenebouaziz: the thing was that when working directly with xarray the code was a lot lot slower, this is why we decided with @RubenImhoff to switch to numpy

hboisgon: I know... it's much faster because you load everything in memory.. and playing more with your chunksize argument / the forcing data chunksize? Because in your case you actually mind less to load full grid for each timestep and would like more much timestep or even all timesteps per grid cell? I know it's faster but I'm afraid for medium basin you may run into trouble already... did you try it for medium/large basins? Else numpy is fine but would be good if for example this still works for basins like the Rhine

hydromt_wflow/workflows/rootzoneclim.py
hboisgon: I know... that's the Delwaq convention but it's not used anywhere else... eg sfincs and setup_riverwidth use discharge. So I think the convention may change anyway so I would say discharge m3/s (and include it in the docstrings!)

hydromt_wflow/wflow.py
laurenebouaziz: very good point, I have changed this, now RootingDepth is no longer an option, but having rootzone_storage in the staticmaps is an option, see comment below

hydromt_wflow/wflow.py
laurenebouaziz: good to have a look later at the auxmaps options! for now I replaced the rootzone_depth flag with a rootzone_storage flag (set as default to False) to only store the RootingDepth maps and no longer the rootzone_storage maps, which are not used by sbm. so this will reduce the amount of new maps in the staticmaps.

hydromt_wflow/workflows/rootzoneclim.py
laurenebouaziz: good point, I have replaced this here and at another point in the code!

laurenebouaziz:
thanks @hboisgon for all the suggestions and quick review! you can see all the new changes in aebd01fbb5581b76a62cb7b9cb19c36e6ccded4d 

(I did not yet run black and merge main to keep the files changed readable)

hydromt_wflow/workflows/rootzoneclim.py
laurenebouaziz: ok! will change this!

hboisgon (approved):

PR #188, opened 2023-06-23
bugfix config updating

### LINKED ISSUE #185: 2D floodplain: missing output.lateral.land in config
This causes an error when building a model from scratch.
Specifically when using the local-inertial option for land routing (2D).

If concerns this piece of code: 
https://github.com/Deltares/hydromt_wflow/blob/main/hydromt_wflow/wflow.py#L653

JoostBuitink:
Checks whether keys are present in config object, before trying to remove them, as this was causing errors. Fixes #185


B-Dalmijn (approved):
Seems to be correct. Better than the try except stuff indeed

PR #190, opened 2023-07-25
Fix example notebooks; update example forcing

JoostBuitink:
The docs had some issues: the example build notebook missed a data catalog (artifact_data), and the plot forcing notebook was no longer working because the forcing files had slightly different coordinate values (floating point issue, likely due to fixing the glacier map issue).


JoostBuitink:
Closing because not a commits show up for some reason

PR #191, opened 2023-07-26
Fix example notebooks; update example forcing; fix docstring formatting

JoostBuitink:
The docs had some issues: the example build notebook missed a data catalog (artifact_data), and the plot forcing notebook was no longer working because the forcing files had slightly different coordinate values (floating point issue, likely due to fixing the glacier map issue). 
Also improved the formatting of some docstrings.


B-Dalmijn (approved):
Most changes were simpel (removed spaces and what not), The larger chunks I checked by looking at the build docs. It looks good.

PR #197, opened 2023-08-11
Workflows

B-Dalmijn:
This is still in progess and some stuff is up for discussion:

- git dependencies vs stable release on conda (HydroMT)
- Which dependencies in the pyproject.toml (Where to rely on HydroMT's dependencies)


.github/workflows/purge_pr_cache.yml
```diff
@@ -0,0 +1,34 @@
...
+          cacheKeysForPR=$(gh actions-cache list -R $REPO -B $BRANCH -L 100 | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
+              gh actions-cache delete $cacheKey -R $REPO -B $BRANCH --confirm
+          done
+          echo "Done"
```
savente93, line 32: I've got a PR open on core that refactors this and significantly improves testing speed so you might want to copy that over once it lands and is working well. For now this is fine though.

.github/workflows/test_dev.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: pip install git+https://github.com/Deltares/hydromt.git@main --user
+
+      - name: Install HydroMT-Wflow plugin
+        run: pip install .
+
+      - name: Run tests
+        run: python -m pytest --verbose --cov=hydromt_wflow --cov-report xml
+
+      - name: Check style
+        run: black --check .
```
savente93, line 74: this should not be necessary here as that is tsted by the pre-commit/lining action

.github/workflows/test_stable.yml
```diff
@@ -0,0 +1,70 @@
...
+        if: steps.cache.outputs.cache-hit != 'true'
+
+      - name: Install HydroMT-Wflow plugin
+        run: pip install .
+
+      - name: Run tests
+        run: python -m pytest --verbose --cov=hydromt_wflow --cov-report xml
+
+      - name: Check style
+        run: black --check .
```
savente93, line 67: this should not be necessary here as that is tsted by the pre-commit/lining action

.github/workflows/test_dev.yml
```diff
@@ -0,0 +1,77 @@
+name: Test dev
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
savente93, line 7: Since these are equal to the triggers on test stable, they are going to always run in parallel. if you want test dev to only happen after stable, I think you need to remove this bit

savente93 (approved):
LGTM, do note that I didnt' review any of the code/docs/config as they all seem linting things, and I don't have knowledge of them. Couple of small comments on the workflows, but aside from that it should be fine

.github/workflows/purge_pr_cache.yml
dalmijn: I'll wait a bit (as you suggested) and create a seperate branch/ PR for it.

.github/workflows/test_dev.yml
dalmijn: Indeed. Was rudimentary code.

.github/workflows/test_stable.yml
dalmijn: Same.

.github/workflows/test_dev.yml
dalmijn: I'll test it. Thanks!

PR #210, opened 2023-10-20
Connect wflow to 1D model

hboisgon:
## Issue addressed
No issue published

## Explanation
This PR introduces a new function to connect wflow to 1D type of model (like Delft3D FM 1D, HEC-RAS, etc.)
In this function a user has two choices to connect:

- derive subbasins based on river network nodes
- derive subbasins based on area. Larger tributaries or upstream boundaries of the river are indeed added as direct gauges rather than subbasins

In this function, gauges can be added and river discharge saved, for the subbasins, two versions are produced, one with all cells and one with river cells only so that wflow can save river.inwater to give the 1D model all water coming into the river (ie the 1D model can then do its own routing and river evaporation).
This has been developed and tested to connect to Delft3D FM for both 1D boundaries and 2D laterals from the subbasins inwater. 

## Checklist


Huite:
Hi @hboisgon,

I saw this come up in my github mentions, so just a drive-by comment.
There's a UGRID1D topology in xugrid: https://deltares.github.io/xugrid/api.html#ugrid1d-topology

The Ugrid1d functionality is rather barebones now, but I think ideally it could help with these kind of workflows. I see several explicit Python loops here (e.g. via `.apply`). Ideally, we could avoid those by making smart use of the Ugrid1d topology and offload most of the work to numba-celltree or something. I'm curious to hear which set of operations you'd require for that to work.

Might be worth some xugrid issues or even a short meeting!

shartgring:
Hi Hélène! I am testing/using this feature for coupling Wflow to a Delft3D FM model. It worked nicely with my geodataframe out of the box with the subbasin approach. However, in order to use the river nodes approach, the geometries have to be ordered to be able to loop them. Or at least, that is how I interpreted the following error that I got:
```
```

I solved this by applying `.explode()` to the geodataframe as suggested here: https://gis.stackexchange.com/questions/378806/multi-part-geometries-do-not-provide-a-coordinate-sequence-error-when-extracti . So, it might be useful to implement .explode() in the hydromt_wflow workflow to prevent these errors. Attached also a zip with the geojson of the geometries that I am using in case you are interested [1D_branches_clipped_20240220.zip](https://github.com/Deltares/hydromt_wflow/files/14342637/1D_branches_clipped_20240220.zip)

hydromt_wflow/workflows/connect.py
```diff
@@ -0,0 +1,283 @@
...
+        # First intersect riv1d with gdf_edges_subbas
+        rivmerge = gpd.overlay(riv1d, gdf_edges_subbas).explode().reset_index(drop=True)
+        # Compute len of river
+        if rivmerge.crs.is_geographic:
+            rivmerge["len"] = rivmerge.geometry.to_crs(3857).length
+        else:
+            rivmerge["len"] = rivmerge.geometry.length
+        # Groupby value and sum length
+        rivmerge = rivmerge.groupby("value")["len"].sum()
+        # Select the subcatch where rivlength is more than 5times rivlen_avg
```
JoostBuitink, line 141: Just a technical question: how sensitive are the results to this factor of 5?

examples/connect_to_1d_model.ipynb
```diff
@@ -0,0 +1,453 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you have a **Wflow** model, one of the applications is that you would like to connect it to a another model such as a 1D hydraulic model like Delft3D or HEC-RAS or a 1D water allocation model like RIBASIM where wflow would give them both river discharge at the boundaries of these models and the runoff generated within the 1D model domain.\n",
+    "\n",
+    "HydroMT-Wflow uses a new function **setup_1dmodel_connection** to try and help you out with these steps. What it does is:\n",
+    "\n",
+    "- Add gauges at the 1D model upstream boundaries to exchange river discharge.\n",
+    "- Derive the sub-basins draining into the 1D model river to exchange the amount of water that enters the river (river inwater).\n",
+    "- Optionnaly, while deriving the sub-basins, large sub-basins can be taken out as tributaries and river discharge is exchanged instead.\n",
```
JoostBuitink, line 20:
```suggestion
    "- Optionally, while deriving the sub-basins, large sub-basins can be taken out as tributaries and river discharge is exchanged instead.\n",
```

JoostBuitink (approved):
Just some small fixes, but really nice work! Great with the example as well!

hydromt_wflow/workflows/connect.py
```diff
@@ -0,0 +1,283 @@
...
+__all__ = ["wflow_1dmodel_connection"]
+
+
+def wflow_1dmodel_connection(
+    gdf_riv: gpd.GeoDataFrame,
+    ds_model: xr.Dataset,
+    connection_method: str = "subbasin_area",
+    area_max: float = 10.0,
+    add_tributaries: bool = True,
+    include_river_boundaries: bool = False,
```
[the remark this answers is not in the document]
hboisgon: A tad complex but to make life easier to post-process wflow netcdf to delft3dfm, boundaries are actual boundaries in the deflt3dfm river while tributaries are point laterals. So actually for delft3dfm with add river boundaries with setup_gauges and laterals (tributaries and subbasins) with this new method:

``` yaml
setup_1dmodel_connection:
  river1d_fn: dfm_river
  connection_method: subbasin_area
  area_max: 2.0
  add_tributaries: True
  include_river_boundaries: False
  mapname: 1dmodel
  update_toml: True
  toml_output: netcdf

setup_gauges:
  gauges_fn: dfm_boundaries
  snap_to_river: True
  snap_uparea: False
  derive_subcatch: False
  basename: 1dboundaries
  toml_output: netcdf
  gauge_toml_header: ["Qb", "Hb"]
  gauge_toml_param: ["lateral.river.q_av","lateral.river.h"]
```

But I could still change the default to True indeed, feels more intuitive.

hboisgon:
> Hi Hélène @hboisgon ! I am testing/using this feature for coupling Wflow to a Delft3D FM model. It worked nicely with my geodataframe out of the box with the subbasin approach. However, in order to use the river nodes approach, the geometries have to be ordered to be able to loop them. Or at least, that is how I interpreted the following error that I got:
> 
> ```
> NotImplementedError: Sub-geometries may have coordinate sequences, but multi-part geometries do not
> ```
> 
> I solved this by applying `.explode()` to the geodataframe as suggested here: https://gis.stackexchange.com/questions/378806/multi-part-geometries-do-not-provide-a-coordinate-sequence-error-when-extracti . So, it might be useful to implement .explode() in the hydromt_wflow workflow to prevent these errors. Attached also a zip with the geojson of the geometries that I am using in case you are interested [1D_branches_clipped_20240220.zip](https://github.com/Deltares/hydromt_wflow/files/14342637/1D_branches_clipped_20240220.zip)

Great, thanks @shartgring for also testing it out and glad the subbasins method worked well directly for your case too :)
Seems like a good idea to add this explode() part so I will!

hydromt_wflow/wflow.py
```diff
@@ -2612,6 +2612,146 @@ def setup_rootzoneclim(
         # update config
         self.set_config("input.vertical.rootingdepth", update_toml_rootingdepth)
 
+    def setup_1dmodel_connection(
+        self,
+        river1d_fn: Union[str, Path, gpd.GeoDataFrame],
+        connection_method: str = "subbasin_area",
+        area_max: float = 10.0,
+        add_tributaries: bool = True,
+        include_river_boundaries: bool = False,
```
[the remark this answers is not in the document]
hboisgon: See above

hydromt_wflow/workflows/connect.py
hboisgon: Good one! I think not too much. So you have the steps where I get the upstream and downstream boundaries of the 1D river and I need to snap these nodes to a wflow river cell before creating the subbasins. So depending on how far we had to snap, some parts of the river 1D lines are still included in the upstream subcatchment that I want to exclude.
So only if we had to snap too far then it may get sensitive. Let's see with practice if this is good enough or if it needs to be refined.

JoostBuitink: Thanks for clarifying, makes sense!

hydromt_wflow/workflows/connect.py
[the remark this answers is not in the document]
JoostBuitink: Ah I see! I agree that True is more intuitive, but good to know that for use-cases as these, you can easily turn them off!

PR #211, opened 2023-10-20
GridModel; HydroMT 0.9 compliant; Deprecated PCR; Ubuntu testing

dalmijn:
## Issue addressed
Fixes #160 

## Explanation
Inherent from GridModel; adjusted methods and properties

## Checklist

Well, the important tests pass, but there is something with testing against core version 0.9 that causes issues...


examples/linux64/wflow_piave_subbasin/log.txt
```diff
@@ -0,0 +1,106 @@
+[ Info: Wflow version `v0.7.2`
+[ Info: Initialize model variables for model type `sbm`.
+[ Info: Cyclic parameters are provided by `staticmaps.nc`.
+[ Info: Forcing parameters are provided by `inmaps.nc`.
+[ Info: Set `vertical.precipitation` using NetCDF variable `precip` as forcing parameter.
+[ Info: Set `vertical.temperature` using NetCDF variable `temp` as forcing parameter.
+[ Info: Set `vertical.potential_evaporation` using NetCDF variable `pet` as forcing parameter.
```
savente93, line 7: Should this log file be included?

.actrc
savente93, line 1: the act files can me useful if you want to test workflows locally but if you don't use act these are not necessary.

examples/linux64/wflow_piave_clip/wflow_sbm.toml
savente93, line 1: Toml :D

examples/linux64/wflow_piave_subbasin/run_results/output.nc
savente93, line 1: In general it is considered bad practice to keep output and/or binaries in the repository, especially if they change often, git cannot deal with blobs as efficently as other types of data. Given they are this small I think it's fine, but if it changes a lot you might want to consider moving these to [git LFS](https://git-lfs.com). I won't repeat it, but the same comment applies to all output files in this PR

hydromt_wflow/wflow.py
```diff
@@ -90,19 +76,32 @@ class WflowModel(Model):
...
+        **artifact_keys,
     ):
+        for lib, version in artifact_keys.items():
+            warnings.warn(
+                "Adding a predefined data catalog as key-word argument is deprecated, "
+                f"add the catalog as '{lib}={version}'"
+                " to the data_libs list instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
savente93, line 93: Given that you also run `super().__init__()` this might lead to duplicated warnings, I'm not sure, but it would be good to double check

hydromt_wflow/wflow.py
```diff
@@ -90,19 +76,32 @@ class WflowModel(Model):
...
+            if not version:  # False or None
+                continue
+            elif isinstance(version, str):
+                lib += f"={version}"
+            data_libs = [lib] + data_libs
```
savente93, line 98: This code also seems to be identical to the one in core. Is it necessary here? I'd expect this to be taken care of by the super call

hydromt_wflow/wflow.py
```diff
@@ -2593,104 +2601,118 @@ def write_staticmaps(self):
...
 
     def write_staticmaps_pcr(self):
         """Write staticmaps at <root/staticmaps> in PCRaster maps format."""
+        warnings.warn(
+            "write_staticmaps_pcr as a method of this object is deprecated. "
+            "Use 'write_staticmaps_pcr' for the 'pcrm' submodule instead. "
+            "Will be removed in hydrom_wflow v0.5.0",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
savente93, line 2649: In some places you use `logger.warning` in others `warnings.warn`, I am not sure if there is a functional difference but unless it is deliberate I think it would be best to try and stay consistent. Especially since `logger.warn` seems to be deprecated but `warnings.warn` not. *shakes fist*

hydromt_wflow/workflows/river.py
```diff
@@ -451,7 +451,7 @@ def river_width(
         values = _precip(ds_like, flwdir=flwdir, logger=logger, **data)
     else:
         if predictor not in ds_like:
-            raise ValueError(f"required {predictor} variable missing in staticmaps.")
+            raise ValueError(f"required {predictor} variable missing in grid.")
```
savente93, line 454: In these situations I find it often helpful to give some indication to which grid it referenced unless that will be clear from context. Up to you whether you want to do anything with this comment.

tests/conftest.py
```diff
@@ -8,8 +9,12 @@
 
 from hydromt_wflow import WflowModel, WflowSedimentModel
 
+SUBDIR = ""
+if platform.system().lower() != "windows":
+    SUBDIR = "linux64"
+
```
savente93, line 15: There does not seem to be a windows folder equivalent, so I wonder if this is necessary? I understand that might be out of scope for this PR but in general it is best to try and make things cross platform so these things are not necessary. I'm happy to help you with this if you want at some point.

tests/conftest.py
```diff
@@ -74,7 +79,7 @@ def clipped_wflow_model():
 @pytest.fixture()
 def floodplain1d_testdata():
     data = xr.load_dataset(
-        join(TESTDATADIR, "floodplain_layers.nc"),
+        join(TESTDATADIR, SUBDIR, "floodplain_layers.nc"),
         lock=False,
         mode="r",
     )
```
savente93, line 85: FYI: fixtures are scoped by module not by test by default in pytest so if you modify this in any way other tests will see that modification. Recently had to find out the hard way. You can add a scope argument to the fixture to change this. Not necessary but wanted to mention it anyway.

.github/workflows/docs.yml
```diff
@@ -1,50 +1,80 @@
...
+      matrix:
+        os: ["ubuntu-latest" ] #, "macos-latest", "windows-latest"]
+        python-version: ['3.11']
+        include:
+        - os: ubuntu-latest
+          label: linux-64
+          prefix: /usr/share/miniconda3/envs/hydromt_wflow
+
+    name: ${{ matrix.label }}
+    runs-on: ${{ matrix.os }}
```
savente93, line 40: if you only test for one version an done os a matrix here is not necessary, but ti does make it easier to add those down the line. up to you whether you want to keep it.

.github/workflows/docs.yml
```diff
@@ -1,50 +1,80 @@
...
 
-      - name: Build docs
+      - name: Generate dummy docs
+        if: ${{ github.event_name == 'pull_request' }}
         run: |
-          pushd docs
-          make html
-          popd
+          export PATH=/usr/share/miniconda3/bin:$PATH
+          PYTHONPYCACHEPREFIX=~/pycache mamba run -n hydromt_wflow sphinx-build ./docs ./docs/_build -b dummy
```
savente93, line 70: You don't seem to use the same caching strategy as core so the export and cache previx is not necessary here. if you want me to explain why this is necessary in core but not here ping me

.github/workflows/test_dev.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           use-mamba: true
 
       - name: Generate env spec
-        run: pip install tomli && python make_env.py test --py-version ${{ matrix.python-version }}
+        run: pip install tomli && python make_env.py test --py-version ${{ matrix.python-version }}.*
```
savente93, line 50: I have a slight susspiscion this is going to break since `.*` is a glob in bash, so it is probably going to expand to things you don't expect. I think the `make_env.py` in core now has a update to deal with this, if not please ping me and I'll make one. If you want to keep this patern you should wrap it in quotes at least

.github/workflows/test_stable.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           use-mamba: true
 
       - name: Generate env spec
-        run: pip install tomli && python make_env.py test --py-version ${{ matrix.python-version }}
+        run: pip install tomli && python make_env.py test --py-version ${{ matrix.python-version }}.*
```
savente93, line 45: I don't think this workflow actually tests against a stable release, you still seem to make the env from the same source, so I think this just duplicates the test dev work, though I'm not 100% on that. Maybe we should have a look at this together one of these days.

savente93 (commented):
I only skimmed the code since I can't really meaninfully review that, but most of it seems fine. I left a couple of comments that you may or may not want to do anything with. The workflows are mostly fine except the stest stable. I'm not sure that one actually does anything (or at least not what you think it does) we might have to pair up and look at that together some time so you can walk me through it. Otherwise LGTM

examples/linux64/wflow_piave_subbasin/log.txt
dalmijn: A case can be made for both I suppose, but given it's testdata that has been produced with HydroMT, I'd say: leave it in there.

.actrc
dalmijn: Agreed, planning on using it from time to time.

examples/linux64/wflow_piave_clip/wflow_sbm.toml
dalmijn: +1

examples/linux64/wflow_piave_subbasin/run_results/output.nc
dalmijn: I agree, it should be something for an overhaul of the tests.

hydromt_wflow/wflow.py
dalmijn: Will look at that.

tests/conftest.py
dalmijn: Also something for a tests overhaul.

.github/workflows/docs.yml
dalmijn: Will do. Will look at that.

.github/workflows/test_stable.yml
dalmijn: We discussed this in person, it does test against dev and stable in the seperate worfkflows.

.github/workflows/docs.yml
dalmijn: I like the fact that it gives me some variables to work with. but en general you're right.

hydromt_wflow/wflow.py
dalmijn: Yes, for backwards compatibility. Its not implemented in the base model class of core, I think. This is/ was done in the DataCatalog I think

JoostBuitink (commented):
Looks good to me! Two things that need fixing, both related to the docs:
- [ ] Update `convert_staticmaps_to_mapstack.ipynb` to reflect the new way to deal with PCRaster files
- [ ] `build_sediment.ipynb` currently fails at the part where we add layers to a wflow model. I don't directly see where it goes wrong, but should hopefully be easy to fix

Perhaps also good to note (in the changelog maybe) that these changes need hydromt v0.9.1 (that is currently not yet released).

savente93 (approved):
LGTM

PR #223, opened 2023-11-23
Update examples installation

### LINKED ISSUE #206: Installation guide not up-to-date

- [X] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt_wflow)

### Kind of issue

Docs are wrong

### Location of the documentation

HydroMT-wflow was simplified with make_env.py but docs were not updated accordingly:

https://deltares.github.io/hydromt_wflow/latest/getting_started/installation.html

https://deltares.github.io/hydromt_wflow/latest/getting_started/example_index.html

https://deltares.github.io/hydromt_wflow/latest/dev_guide/dev_install.html

The second one is also not clear because if people install from conda-forge plotting dependencies as well as jupyter lab wont be installed.

The third link, how to install make is not mentioned.

### Documentation problem

As of now people cannot install hydromt-wflow with the recommended install

### Suggested fix for documentation

Update with steps from make_env.py

hboisgon:
## Issue addressed
Fixes #206 

## Explanation
I think the installation was already updated but the examples not fully (installing jupyter and downloading the correct version of the examples).
I also do not think we need matplotlib-base in the [examples] install list as matplotlib is already installed and matplotlib-base is not installable with pip.

## Checklist

## Additional notes
In hydromt core or the delft3dfm plugin, I recommend installing from pypi like so:
pip install "hydromt[slim]"
This avoids that people need to manually add libraries in [extra] and [examples] one by one or clone the github repository.
It does not work at the moment because of the matplolib-base library in [examples] (which I now removed) and the pcraster library in [extra]. 
I think an easy way to install and run the examples would be nice. For this, maybe we could move the pcraster library in another list than extra? Else we could also add gwwapi, hydroengine and pyet in the examples list instead?

### DIFF
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -20,16 +20,65 @@ The first step is to install HydroMT-Wflow and the other Python dependencies in 
 see the **Install HydroMT-Wflow in a new environment** section in the
 `installation guide <https://deltares.github.io/hydromt_wflow/latest/getting_started/installation.html>`_
 
+To run the notebooks, you need to install additional dependencies that are jupyterlab to
+run the notebooks and cartopy for some plots. To install these packages in your existing
+hydromt-wflow environment, do (these packages are also available in conda-forge):
 
-2 - Download the content of the HydroMT github repository
-*********************************************************
-To run the examples locally, you will need to download the content of the examples folder in the HydroMT-Wflow repository.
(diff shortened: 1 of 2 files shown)


JoostBuitink (approved):
Thanks for fixing this, LGTM!

PR #229, opened 2024-01-11
Implement NoDataStrategy to deal with optional setup_functions

### LINKED ISSUE #208: Compatibility with nodata requirements of core
### Kind of request

### Enhancement Description

The current version of hydromt_wflow does not work with core 0.9 because of how missing data is handled in the `DataAdapters`. I propose a new method here: https://github.com/Deltares/hydromt/issues/584. Implementing this with the current behaviour of the wflow plugin (ignore) could be as simple as adding a flag to ignore missing data (like reservoirs) or to raise if the component is critical for the model (like rainfall).

### Use case

If I want to run with the main branch of HydroMT (in this case for the cloud pilot), this feature should be implemented.

### Additional Context

_No response_

JoostBuitink:
## Issue addressed
Fixes #208, but needs https://github.com/Deltares/hydromt/issues/731 to be fixed in order to work.

## Explanation
Implemented a NoDataStrategy that ignores errors for the optional setup function (lakes, reservoirs, glacier and gauges). Now a warning is returned (in stead of an error), similar to the old behavior of hydromt_wflow. However, as the NoDataStrategy is currently incomplete in HydroMT-core, we need https://github.com/Deltares/hydromt/issues/731 to be fixed before this fix works.

## Checklist


JoostBuitink:
Stable tests are failing, because I believe that they do no yet use the latest version of hydromt (v0.9.3).

hydromt_wflow/wflow.py
```diff
@@ -1677,6 +1692,8 @@ def setup_reservoirs(
 
             for option in res_toml:
                 self.set_config(option, res_toml[option])
+        else:
+            self.logger.info("Skipping method, as no data has been found")
```
dalmijn, line 1696: I'd say, for consistency's sake, make this the same as the other methods, i.e.:
```Python
if ... is None:
    logger.warning(< warning message >)
    return
```

dalmijn (changes requested):
Thanks for the good work! LGTM. I'd only request that `setup_reservoirs` is structured the same in this regard as the other methods.

Starphire1:
I'm excited about a new release. It would be great to have water level management and inflow/outflow calculations for lakes/reservoirs (if possible). It would also be nice if the created toml file at model build stage includes, e.g. lake = "wflow_lakelocs" and reservoir = "wflow_reservoirlocs" to reduce a manual model build step. The reservoir model build has been an exciting new addition to the model. Thanks for all the updates!

dalmijn (approved):
LGTM.

PR #230, opened 2024-01-19
update config from ini to yml

### LINKED ISSUE #207: Config ini files should be translated to yml

### Reproducible Example

NA

### Current behaviour

The wflow config models are currently still in ini format which is deprecated in hydromt-core and will be removed in the 1.0 release.  

### Desired behaviour

To maintain good support the current config files should be translated to a supported file format such as toml or yaml.

### Additional context

_No response_

hboisgon:
## Issue addressed
Fixes #207 

## Explanation
Update hydromt config files from ini to yaml. Also updated the tests and docs.

## Checklist


hboisgon:
I removed the _archive folder. There were still some old docs files on wflow build/update etc. I don't think we need to track this so I decided to remove rather than update the files.

dalmijn (approved):
Everything checks out. 

There seems to be a docs bug, but I made a separate issue for that #233 

Also, the docstrings for the base 'Model' object methods 'update' and 'build' should be updated at some point.

PR #247, opened 2024-02-20
Several basin related fixes

### LINKED ISSUE #245: Wrong dtype for wflow_subcatch map

### Reproducible Example

Happens if these conditions are met:

- building a model with kind basin
- basin variable not available in hydrography_fn or available with wrong dtype
- no upscaling required (model resolution is the same as hydrography_fn)

### Current behaviour

The output wflow_subcatch dtype can be np.uint32 instead of np.int32. Not a problem to run wflow but for example vectorizing the map does not work anymore (which is used in the ``read_results`` method):

```
model.grid["wflow_subcatch"].raster.vectorize()

> ValueError: image dtype must be one of: int16, int32, uint8, uint16, float32

``` 

In case of basin variable not available in hydrography_fn, the uint32 actually pops up in the ``hydromt.workflows.get_basin_geometry``:

https://github.com/Deltares/hydromt/blob/d634534fd1c3c5e470d6835a1ee1c518705576ad/hydromt/workflows/basin_mask.py#L380

### Desired behaviour

Make sure that basin dtype is np.int32 in that part of the ``hyrography`` workflow:

https://github.com/Deltares/hydromt_wflow/blob/c2074330ea28fae65bd4f184e4c9bf8405cbf117/hydromt_wflow/workflows/basemaps.py#L204-L216

### Additional context

_No response_

### LINKED ISSUE #246: In setup_outlets use the same IDs as the wflow_subcatch map
### Kind of request

### Enhancement Description

Right now the wflow basins in wflow_subcatch and the outlets in wflow_gauges IDs do not match (outlet of basin 1 does not automatically have ID 1).

### Use case

This is nice to avoid confusion when post-processing wflow model results if the IDs of the outlets can match else this becomes quite some post-processing work to match the right outlet with the right basin.

### Additional Context

_No response_

hboisgon:
## Issue addressed
Fixes #245 
Fixes #246 

## Explanation
For #245, seems there was only one case subcatch dtype was not np.int32. Rather than fixing the hydromt get_basin_geometry workflow I chose to force np.int32 dtype so that a different basins dtype in local hydrography would then also be supported.

For #246, just use the wflow_subcatch ID. I chose to just have this happen all the time rather than add an user option to still get random IDs.

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,9 +14,11 @@ Added
 
 Changed
 -------
+- **setup_outlets**: the IDs of wflow_subcatch are used to define the outlets IDs rather than [1:n]. PR #247
 
 Fixed
 -----
+- Wrong dtype for wflow_subcatch map. PR #247
 
 v0.5.0 (February 2024)
 ======================
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1047,6 +1047,11 @@ skipping adding gauge specific outputs to the toml."
     ):
         """Set the default gauge map based on basin outlets.
 
+        If wflow_subcatch is available, the catchment outlets IDs will be matching the
+        wflow_subcatch IDs. If not, then IDs from 1 to number of outlets are used.
+
+        Can also add csv/netcdf output settings in the TOML.
+
         Adds model layers:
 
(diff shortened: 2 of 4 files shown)


dalmijn (approved):
LGTM.

PR #248, opened 2024-02-20
Snapping to river with uparea method

### LINKED ISSUE #224: Gauge snap_uparea can snap to non-river pixels

### Reproducible Example

For example, some stations in the GRDC database have very small values for uparea (<30 km2). Using the snap_uparea method, these gauges will be snapped to the pixel with the closest value for uparea, but that does not mean that it is a river pixel

### Current behaviour

Gauges with small upareas are correctly snapped, but this will give an error when running wflow when you request river discharge as output.

### Desired behaviour

As this feature is mostly relevant for discharge stations (I think), it would better if only the river cells are considered to snap towards. Alternatively, gauges with an uparea smaller than the uparea used in `setup_rivers` could also be ignored. But I think the first option is better. 

### Additional context

_No response_

hboisgon:
## Issue addressed
Fixes #224 

## Explanation
Allow to pass a mask to the ``gauge_map_uparea`` workflow.
Also found a bug that validation was not properly done (should be both validation on relative and absolute error).

## Checklist


hydromt_wflow/workflows/gauges.py
```diff
@@ -61,13 +64,33 @@ def gauge_map_uparea(
...
 
+    # Find if there are any nodata in gdf["uparea"]
+    if gdf["uparea"].isna().any():
+        logger.warning(
+            "There are NaN values in the gauges uparea."
+            "Uparea from wflow will be used ie no snapping if mask is None."
+        )
+        uparea_wflow = ds[uparea_name].raster.sample(gdf, wdw=0)
+        # Replace nan values in gdf with uparea from wflow
+        gdf["uparea"] = gdf["uparea"].fillna(uparea_wflow.to_pandas())
```
dalmijn, line 75: Was wondering, I can supply an empty column `uparea` here and it will fill out all the values.
Shouldnt we let it pass even if `uparea` is not present in the vector file and just fill it in as one has to deliberately set snap_uparea to `True`. We can add a warning that the column was not found and one will be created (even if only in memory).

hydromt_wflow/workflows/gauges.py
```diff
@@ -61,13 +64,33 @@ def gauge_map_uparea(
...
+        )
     logger.debug(
         f"Snapping gauges points to best matching uparea cell within wdw (size={wdw})."
     )
     upa0 = xr.DataArray(gdf["uparea"], dims=("index"))
     upa_dff = np.abs(ds_wdw[uparea_name].where(ds_wdw[uparea_name] > 0).load() - upa0)
-    upa_check = np.logical_or((upa_dff / upa0) <= rel_error, upa_dff <= abs_error)
+    upa_check = np.logical_and((upa_dff / upa0) <= rel_error, upa_dff <= abs_error)
     # find best matching uparea cell in window
     i_wdw = upa_dff.argmin("wdw").load()
```
dalmijn, line 96: I have a test case where one gauges does not fall within the range where there is a search for a river. This causes all `NaN` values in 'upa_dff' for a certain index. This causes a crash when trying to apply `argmin` to 'upa_dff'.

We should I guess account for gauges that are unrealistic (in terms of where the rivers are) and out of reach of the rivers.

dalmijn (changes requested):
@hboisgon, Thanks for the changes! The logic seems sound.
However I got stopped half-way by an error that arose from a test model I use.

This will therefore a bit of a premature review as the program halts.

hboisgon:
Thanks for the review @dalmijn. Can you please send your test? This will be easier to debug.

hydromt_wflow/workflows/gauges.py
hboisgon: I don't think it's good if no ``uparea`` column is present to still use this method because the concepts are really different.

If you want to snap to the river then we use the flow directions to find the closest dowsntream river cell.

If you want to snap with uparea, we select a window of cells around your point, and optionally mask in this window valid river cells. But because if your point is not already in a river then ``uparea`` can be super low (ie 1 or 2 cells uparea) then most chances are that you will never match the ``uparea`` criteria to be in a river, as uparea classically is more than 30 km2 for river cells.
So if you have no ``uparea`` column and want to snap to river this will most probably go wrong unless your points are already in the right locations. So better use the classic snapping to river technique.

So actually you could argue that I should here just skip points where `uparea` is NaN instead of still trying in case mask is not None and the points are not already in the mask.

dalmijn: I get your reasoning, but e.g. actual gauge location (with unknown upareas) will be in the right location (I'd presume). From that standpoint I'd rather still would be able to run this snap_uparea. However, in that case, I do not see the difference between a completely empty column and an absent one.

JoostBuitink: > I'd rather still would be able to run this snap_uparea

Briefly glancing through this discussion, and I don't really follow this comment. If no uparea is provided for a gauge, how would we be able to still do this uparea snapping?

dalmijn: After discussing a bit with @JoostBuitink, it seems indeed that in the context of this method, it would make more sense to delete those that do not have an uparea value. I am however opposed to the idea of discarding these gauges all together.

hboisgon: So let's take an example where 10 gauges are at the right locations but one is not. Then for a user is is easier to add upstream area of that one gauge only instead of all of them that are in the right place already. Which is what the current implementation tries to do.

If no uparea column this means no snapping or you don't need to snap so just do not allow snapping in the setup_gauges method.

@JoostBuitink do you see a use case for the first example (user could also just change the coordinates and move the point instead of updating uparea properties). If not then indeed I can change and I think actually go back to not filling uparea in case of NaN.

JoostBuitink: I do see that use-case, what got me thinking is the following line-of-reasoning: we do the snapping with the idea that the location of the actual gauges (as provided in the shapefile) does not correspond with the locations where they should end up inside the wflow model. Using the uparea information, we can now better find the wflow-pixel that matches the uparea best with the gauge itself. 

However, if this information is not present, getting the uparea value from the wflow model sounds a bit counterintuitive, since we suspect that location of the gauge and the wflow-pixel do not match. Therefore I am not sure whether using the (wflow-pixel) uparea value to do the uparea_snapping makes sense. 

I don't really have a strong preference for one of the two (filling uparea, or skipping these gauges), but I do think it is important to show a clear logging message that we either ignore this gauge, or fill the uparea value with the wflow-uparea value. And maybe log which gauges (ids) are affected by this.

hboisgon: I decided to add a flag (False by default) to fill upstream area values with wflow values. Also I hope clearer warning messages on which gauges have NaN and if they are skipped or included with no snapping. And an extra message on the number of gauges snapped compared to total number of gauges in input data.
I hope it works like this. It's not too much work to support both and might be nice for big datasets like GRDC in case there are a few NaNs for some gauges that they are not skipped if they already in the right spot :)

hydromt_wflow/workflows/gauges.py
hboisgon: There was indeed a bug here that should be now fixed! Your example still does not work because I added a check if all gauges have uparea==NaN as users should either use no snapping or in your case the default snap_to_river method rather than snap_uparea.

hboisgon:
@dalmijn and @JoostBuitink I corrected the bugs and added a flag for fillna in upstream are that is false by default (nodata gauges are skipped). So that the users knows what he is doing if he decides to put the flag to True :)
Hope the PR is now ready to be merged!

hydromt_wflow/workflows/gauges.py
JoostBuitink: Nice :D! Adding clear logging messages and a flag to let the user decide how to deal with those is indeed the best solution! Thanks a lot for addressing this!

dalmijn (approved):
Looked through the changes and tested it. 
LGTM!

PR #252, opened 2024-03-06
States

### LINKED ISSUE #18: Implement wflow states object
Implement a states object for wflow:
- read / write
- initialize with default values

hboisgon:
## Issue addressed
Fixes #18 

## Explanation
Added a method to prepare cold states for wflow.
To be precise as this can be done after a wflow model has been updated and calibrated, I added a new function to get the right values based on value/scale/offset options in the config file with a new utils.get_grid_from_config function.

## Checklist

### DIFF
--- a/hydromt_wflow/utils.py
+++ b/hydromt_wflow/utils.py
@@ -1,13 +1,15 @@
 """Some utilities from the Wflow plugin."""
+from os.path import abspath, join
 from pathlib import Path
-from typing import Dict, Union
+from typing import Dict, Optional, Union
 
 import numpy as np
 import xarray as xr
 from hydromt.io import open_timeseries_from_table
 from hydromt.vector import GeoDataArray
+from hydromt.workflows.grid import grid_from_constant
 
-__all__ = ["read_csv_results"]
+__all__ = ["read_csv_results", "get_config", "get_grid_from_config"]
 
 
 def read_csv_results(fn: Union[str, Path], config: Dict, maps: xr.Dataset) -> Dict:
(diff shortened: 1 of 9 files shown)


dalmijn (changes requested):
@hboisgon, Thanks for the good work! For the most part, it looks very solid!

I had a few points:

-  `path_input` is not set (could cause issues with custom config files)
-  Same argument goes for the states themselves
( I realise we rely upon the blueprint settings files, but this makes it a bit more future proof)
- Nodata value of `waterlevel_lake`
- A bit more documentation maybe in the building (or updating) model notebooks that this can be done.  I'd say it's really key that the states can be set like this.

hydromt_wflow/utils.py
```diff
@@ -170,3 +172,144 @@ def read_csv_results(fn: Union[str, Path], config: Dict, maps: xr.Dataset) -> Di
...
     return csv_dict
+
+
+def get_config(
+    *args,
+    config: Dict = {},
+    fallback=None,
+    root: Path = None,
+    abs_path: Optional[bool] = False,
+):
```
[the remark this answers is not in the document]
hboisgon: As discussed I don't really like to pass the whole WflowModel in a workflow. This method in a way is just recursive look into a dict and includes the default value to return if not found so it's also not too complex. Let's re-evaluate with hydromt v1 when ModelComponent like config will be separated from Model.

hydromt_wflow/wflow.py
```diff
@@ -3637,3 +3726,27 @@ def clip_forcing(self, crs=4326, **kwargs):
...
+            # Check for reservoirs/lakes presence in the clipped model
+            remove_maps = []
+            if self._MAPS["resareas"] not in self.grid:
+                if "volume_reservoir" in ds_states:
+                    remove_maps.extend(["volume_reservoir"])
+            if self._MAPS["lakeareas"] not in self.grid:
+                if "waterlevel_lake" in ds_states:
+                    remove_maps.extend(["waterlevel_lake"])
+            ds_states = ds_states.drop_vars(remove_maps)
+            self.set_states(ds_states)
```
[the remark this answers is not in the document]
hboisgon: I'm not sure I understand your comment? You mean the xr.merge?
So yes it's because in core it's a dict and not direct xr.Dataset. But I can either merge or go through the dictionnary keys one by one before clipping too.
Part of the method could definitely be in core yes and there is an issie for it to have general clip methods. But anyhow here it's really a couple of lines and the remove_maps part is specific to wflow so I need to also add the function in hydromt_wflow anyway.

hboisgon:
Thanks for the review @dalmijn ! I addressed most of comments. Could you check again?
Some extra replies:

> @hboisgon, Thanks for the good work! For the most part, it looks very solid!
> 
> I had a few points:
> 
> * `path_input` is not set (could cause issues with custom config files)

This is handled in the ``read_states`` and ``write_states`` methods normally :)

> * Same argument goes for the states themselves
>   ( I realise we rely upon the blueprint settings files, but this makes it a bit more future proof)

I added.

> * Nodata value of `waterlevel_lake`

Updated. 

> * A bit more documentation maybe in the building (or updating) model notebooks that this can be done.  I'd say it's really key that the states can be set like this.

The only use case for this I know is using wflow in Delft-FEWS because Wflow.jl can setup its own cold states so you don't want to prepare them by default (so should not be included in the wflow_build.yml template). The use is also not really complex so not sure if I need a specific update example for it. It's already documented in the list of available methods in the wflow docs and properly listed in the API. Do you think we really need to add an example?

hydromt_wflow/wflow.py
```diff
@@ -3274,17 +3336,47 @@ def write_forcing(
...
+
+        if self.states:
+            self.logger.info("Writting states file")
+
+            # get output filename
+            if fn_out is not None:
+                self.set_config("state.path_input", fn_out)
+                self.write_config()  # re-write config
+            else:
+                fn_out = self.get_config("state.path_input", abs_path=True)
```
dalmijn, line 3366: A fall back is necessary here I think, something this maybe?:
```python
                fn_out = self.get_config(
                    "state.path_input", 
                    fallback=Path(
                        self.root,
                        "instate",
                        "instates.nc",
                    ),
                    abs_path=True,
                )
```

dalmijn (changes requested):
Thanks @hboisgon for addressing the comments!

Everything looks good except for one:

> > @hboisgon, Thanks for the good work! For the most part, it looks very solid!
> > I had a few points:
> > 
> > * `path_input` is not set (could cause issues with custom config files)
> 
> This is handled in the `read_states` and `write_states` methods normally :)
> 

This is not handled properly in write_states. `state.path_input` will not be set.
An error will be thrown in `write_states` as there is no value for it in the config. 

> > A bit more documentation maybe in the building (or updating) model notebooks that this can be done. I'd say it's really 
> > key that the states can be set like this.
>
> The only use case for this I know is using wflow in Delft-FEWS because Wflow.jl can setup its own cold states so you don't 
> want to prepare them by default (so should not be included in the wflow_build.yml template). The use is also not really 
> complex so not sure if I need a specific update example for it. It's already documented in the list of available methods in the > wflow docs and properly listed in the API. Do you think we really need to add an example?
> 

I agree, I was more thinking of a added cell in the current update or build notebook, not a separate one.

hydromt_wflow/wflow.py
hboisgon: Good catch! Solved it now.

hydromt_wflow/wflow.py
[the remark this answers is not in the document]
dalmijn: Alright, good to know!

hydromt_wflow/wflow.py
```diff
@@ -3274,17 +3336,50 @@ def write_forcing(
...
+
+            # get output filename
+            fn_default = join(self.root, "instate", "instates.nc")
+            if fn_out is not None:
+                self.set_config("state.path_input", fn_out)
+                self.write_config()  # re-write config
+            else:
+                fn_out = self.get_config(
+                    "state.path_input", abs_path=True, fallback=fn_default
+                )
```
dalmijn, line 3369: The fallback is nice, but the config file is not set with the path if this `else` statement is hit. This mean that `state.path_input` is not set.

However even if we do this, the config is written (`WflowModel.write_config`) before the states are written in `WflowModel.write`.  I guess we either have to catch this in `setup_cold_states`, or write the config file at a later stage in `WflowModel.write`.

dalmijn (changes requested):
@hboisgon, Thanks for the update :)!

I have one last point remaining.
- setting `state.path_input`

hydromt_wflow/wflow.py
hboisgon: Indeed! Hopefully fixed now

dalmijn (approved):
LGTM.

@hboisgon could you take a quick look at the `make_path_relative` functionm?

hboisgon:
Hi @dalmijn. In the end I preferred to handle this similarly to other components like forcing.
While checking forcing and grid, I realized that wflow also supports dir_input as relative path to all input files including states so I added that as well.
In the end I do not think we need the function to deal with relative path that you added (even if it was a good one to check for mount first). If you agree then we can remove it I think.

dalmijn:
@hboisgon, It look good to me. I'll delete the excess functions.

PR #255, opened 2024-03-14
Setup KsatHorFrac

### LINKED ISSUE #228: Include KsatHorFrac estimation method from Awad Ali's MSc thesis
### Kind of request

Adding new functionality

### Enhancement Description

In his MSc thesis, Awad Ali tested Random Forest (RF) and Boosted Regression Trees (BRT) methods to estimate the KsatHorFrac parameter of wflow_sbm. The methods show good results and may be a first step to an alternative for the now uniform KsatHorFrac parameter (with value 100). Awad has already prepared two global maps (one based on RF and one based on the BRT algorithm), which can be found on the wflow P-drive (`hydromt_staging/kstathorfrac_thesis_awadali/`). Because he trained the RF and BRT models on the CAMELS-UK dataset in the UK, the method may not yet be working well in regions with different geology and soil types. But this is a point of further research at the moment.

I think the results are ready for usage in hydromt as an alternative option to the standard uniform KsatHorFrac derivation. In the additional context below, I've provide a function that we use to include one of the KsatHorFrac maps to update an existing wflow_sbm model. Let's see if we can include this in hydromt_wflow in such a way that we can do this up front through a function. Can you assist me in creating a hydromt function that can make use of other KsatHorFrac maps? I am not familiar enough with all functions and procedures in the current version of hydromt to directly open a PR myself, so let's first see where the function belongs and what it should look like. If we make it general enough, you can switch between the 'standard' derivation, the RF and BRT methods (or possibly something else in the future). I can imagine that we update the RF and BRT maps at some point based on future research, but that shouldn't change the hydromt functionality, I assume. 

### Use case

The KsatHorFrac parameter is currently derived as a uniform parameter with value 100, which has to be calibrated later on for improved model performance. Some sort of transfer function for this parameter would be ideal. With Awad's work, we have maps of gridded KsatHorFrac values that can be used in wflow_sbm. These values are based on a ML-based transfer functions between soilgrids data and the KsatHorFrac value. With the current global maps, we expect good results in regions with geology and soil types similar to the UK. For other regions, this has still to be tested.

### Additional Context

Below the function we use to update the wflow model with the new KSatHorFrac maps. The maps are stored at the soilgrids spatial resolution and are store as log10(KSatHorFrac). 

```python
import os
import numpy as np

import hydromt
from hydromt_wflow import WflowModel


# A function adjust_ksathorfrac is defined that takes in a model_folder path. 
# The function adjusts the KsatHorFrac parameter grid based on either a 
# random forest or boosted regression trees based global parameter map.
def adjust_ksathorfrac_and_toml_file(
    model_folder, 
    ksathorfrac_path=None, 
    pred_method=None,
    ):
    """
    Adjust the KsatHorFrac parameter grid with empirical relationship.

    Args:
        model_folder (str): The path to the wflow model.
        ksathorfrac_path (str): optional. the path to the .tif file with the global KsatHorFrac 
                                maps based on the random forest or boosted regression trees
                                methods. Standard set to None.
        pred_method (str): optional. 'RF' for Random Forest, and 'BRT' for Boosted Regression 
                                Trees. If None, the KsatHorFrac value will not be changed. 
                                Standard set to None.

    Returns:
        None.    
    """
    # Open the wflow model
    print(f"Reading wflow model at {model_folder}")
    mod = WflowModel(model_folder, config_fn="wflow_sbm.toml", mode="r+")
    
    # Change KsatHorFrac if requested
    if pred_method is not None:
        if pred_method == "RF" or pred_method == "BRT":
            var_name = pred_method + '_250_KsatHorFrac_global'
        
            # read (clipped by extent) KsatHorFrac with data_catalog.get_rasterdataset
            da_KsatHorFrac = mod.data_catalog.get_rasterdataset(
                data_like = ksathorfrac_path,
                geom = mod.staticgeoms['basins']
                )
            
            # tranform to a logarithmic scale
            da_KsatHorFrac.values = np.log10(da_KsatHorFrac.values)

            # resample KsatHorFrac with grid_from_rasterdataset
            resampled_KsatHorFrac = hydromt.workflows.grid.grid_from_rasterdataset(
                grid_like = mod.staticmaps['KsatHorFrac'],
                ds = da_KsatHorFrac,
                reproject_method = ['average'])
            
            # tranform back the log KsatHorFrac
            resampled_KsatHorFrac[var_name] = 10**resampled_KsatHorFrac[var_name]
            
            # fill missing values
            filled_KsatHorFrac = resampled_KsatHorFrac
            filled_KsatHorFrac[var_name] = filled_KsatHorFrac[var_name].interpolate_na(dim = 'lon', method='linear')

            # fill NA's around the boundary
            filled_KsatHorFrac[var_name] = filled_KsatHorFrac[var_name].interpolate_na(dim = 'lon', method='nearest', fill_value="extrapolate")

            # add and write to staticmap
            mod.set_staticmaps(filled_KsatHorFrac[var_name], 
                            name="KsatHorFrac_"+pred_method)

    # Edit model config
    setting_toml = {}
    if pred_method == "RF":
        setting_toml["input.lateral.subsurface.ksathorfrac"] = "KsatHorFrac_RF"
    elif pred_method == "BRT":
        setting_toml["input.lateral.subsurface.ksathorfrac"] = "KsatHorFrac_BRT"

    # Loop through each setting defined in setting_toml and update 
    # it in the model configuration
    for option in setting_toml:
        mod.set_config(option, setting_toml[option])

    #Write staticmaps and new TOML config
    if pred_method is not None:
        mod.write_staticmaps()
    mod.write_config(config_name="wflow_sbm_calibrated.toml")

    return None
```

dalmijn:
## Issue addressed
Fixes #228 

## Explanation
Raw implementation of the setup from #249 

## Checklist

No.

### DIFF
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1925,8 +1925,8 @@ or created by a third party/ individual.
         ksat_fn : str, optional
             The identifier of the KsatHorFrac dataset in the data catalog.
         variable : str | None, optional
-            The identifier of the method via which this version of the KsatHorFrac \
-dataset was created (as multiple can exist), by default None.
+            The variable name for the ksathorfrac map to use in ``ksat_fn`` in case \
+``ksat_fn`` contains several variables. By default None.
         resampling_method : str, optional
             The resampling method when up- or downscaled, by default "average"
         """
@@ -1943,13 +1943,12 @@ dataset was created (as multiple can exist), by default None.
         # Ensure its a DataArray
         if isinstance(dain, xr.Dataset):
             raise ValueError(
-                f"The ksathorfrac data should be \
-in a xarray.DataArray format; it's in a {type(dain).__name__} format. \
-This could indicate the absence of a correct 'variable' value."
+                "The ksathorfrac data contains several variables. \
+Select the variable to use for ksathorfrac using 'variable' argument."
             )
 
         # Create scaled ksathorfrac map
-        dsout = workflows.ksathorfrac(
+        daout = workflows.ksathorfrac(
             dain,
             ds_like=self.grid,
             resampling_method=resampling_method,
@@ -1966,7 +1965,7 @@ This could indicate the absence of a correct 'variable' value."
             lname += f"_{variable}"
 
         # Set the grid
-        self.set_grid(dsout, name=lname)  # TODO of course
+        self.set_grid(daout, name=lname)
         self.set_config("input.lateral.subsurface.ksathorfrac", lname)
 
     def setup_glaciers(self, glaciers_fn="rgi", min_area=1):
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1932,17 +1932,13 @@ dataset was created (as multiple can exist), by default None.
         """
         self.logger.info("Preparing KsatHorFrac parameter map.")
 
-        if isinstance(ksat_fn, str):
-            # Get the data from the catalog
-            dain = self.data_catalog.get_rasterdataset(
-                ksat_fn,
-                geom=self.region,
-                buffer=2,
-                variables=variable,
-                single_var_as_array=True,
-            )
-        else:
-            dain = ksat_fn.copy()
+        dain = self.data_catalog.get_rasterdataset(
+            ksat_fn,
+            geom=self.region,
+            buffer=2,
+            variables=variable,
+            single_var_as_array=True,
+        )
 
         # Ensure its a DataArray
         if isinstance(dain, xr.Dataset):
@@ -1959,9 +1955,19 @@ This could indicate the absence of a correct 'variable' value."
             resampling_method=resampling_method,
         )
 
+        # Set the output variable name
+        if not isinstance(ksat_fn, str):
+            bname = ksat_fn.name if ksat_fn.name is not None else "KsatHorFrac"
+        else:
+            bname = ksat_fn  # base name of the outgoing layer name
+
+        lname = bname
+        if variable is not None:
+            lname += f"_{variable}"
+
         # Set the grid
-        self.set_grid(dsout)  # TODO of course
-        self.set_config("input.lateral.subsurface.ksathorfrac", "KsatHorFrac")
+        self.set_grid(dsout, name=lname)  # TODO of course
+        self.set_config("input.lateral.subsurface.ksathorfrac", lname)
 
     def setup_glaciers(self, glaciers_fn="rgi", min_area=1):
         """
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -1908,6 +1908,61 @@ a map for each of the wflow_sbm soil layers (n in total)
         # Update the toml file
         self.set_config("model.thicknesslayers", wflow_thicknesslayers)
 
+    def setup_ksathorfrac(
+        self,
+        ksat_fn: Union[str, xr.DataArray],
+        variable: str | None = None,
+        resampling_method: str = "average",
+    ):
+        """Set KsatHorFrac parameter values from a predetermined map.
+
+        This predetermined map contains (preferably) 'calibrated' values of \
+the KsatHorFrac parameter. This map is either selected from the wflow Deltares data \
+or created by a third party/ individual.
+
+        Parameters
+        ----------
+        ksat_fn : str, optional
+            The identifier of the KsatHorFrac dataset in the data catalog.
+        variable : str | None, optional
+            The identifier of the method via which this version of the KsatHorFrac \
+dataset was created (as multiple can exist), by default None.
+        resampling_method : str, optional
+            The resampling method when up- or downscaled, by default "average"
+        """
+        self.logger.info("Preparing KsatHorFrac parameter map.")
+
+        if isinstance(ksat_fn, str):
+            # Get the data from the catalog
+            dain = self.data_catalog.get_rasterdataset(
+                ksat_fn,
+                geom=self.region,
+                buffer=2,
+                variables=variable,
+                single_var_as_array=True,
+            )
+        else:
+            dain = ksat_fn.copy()
+
+        # Ensure its a DataArray
+        if isinstance(dain, xr.Dataset):
+            raise ValueError(
+                f"The ksathorfrac data should be \
+in a xarray.DataArray format; it's in a {type(dain).__name__} format. \
+This could indicate the absence of a correct 'variable' value."
+            )
+
+        # Create scaled ksathorfrac map
+        dsout = workflows.ksathorfrac(
+            dain,
+            ds_like=self.grid,
+            resampling_method=resampling_method,
+        )
+
+        # Set the grid
+        self.set_grid(dsout)  # TODO of course
+        self.set_config("input.lateral.subsurface.ksathorfrac", "KsatHorFrac")
+
     def setup_glaciers(self, glaciers_fn="rgi", min_area=1):
         """
         Generate maps of glacier areas, area fraction and volume fraction.
--- a/hydromt_wflow/workflows/soilparams.py
+++ b/hydromt_wflow/workflows/soilparams.py
@@ -1,0 +1,53 @@
+"""Functions for individual soil parameters."""
+
+import numpy as np
+import xarray as xr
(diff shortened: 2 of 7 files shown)


hydromt_wflow/wflow.py
```diff
@@ -1908,6 +1908,61 @@ def setup_soilmaps(
...
+            # Get the data from the catalog
+            dain = self.data_catalog.get_rasterdataset(
+                ksat_fn,
+                geom=self.region,
+                buffer=2,
+                variables=variable,
+                single_var_as_array=True,
+            )
+        else:
+            dain = ksat_fn.copy()
```
hboisgon, line 1945:
```suggestion
```
Not needed, if a xr.DataArray is passed, get_rasterdataset takes that into account and can still do things like slicing for example.

hydromt_wflow/wflow.py
```diff
@@ -1908,6 +1908,61 @@ def setup_soilmaps(
...
+the KsatHorFrac parameter. This map is either selected from the wflow Deltares data \
+or created by a third party/ individual.
+
+        Parameters
+        ----------
+        ksat_fn : str, optional
+            The identifier of the KsatHorFrac dataset in the data catalog.
+        variable : str | None, optional
+            The identifier of the method via which this version of the KsatHorFrac \
+dataset was created (as multiple can exist), by default None.
```
hboisgon, line 1929:
```suggestion
            The variable name for the ksathorfrac map to use in ``ksat_fn`` in case ``ksat_fn`` contains several variables.
```

hydromt_wflow/wflow.py
```diff
@@ -1908,6 +1908,61 @@ def setup_soilmaps(
...
+        else:
+            dain = ksat_fn.copy()
+
+        # Ensure its a DataArray
+        if isinstance(dain, xr.Dataset):
+            raise ValueError(
+                f"The ksathorfrac data should be \
+in a xarray.DataArray format; it's in a {type(dain).__name__} format. \
+This could indicate the absence of a correct 'variable' value."
+            )
```
hboisgon, line 1953: Nice to do an error message here :) Because you test on xr.Dataset, then you can simplify the error message fort he user I think.
```suggestion
        if isinstance(dain, xr.Dataset):
            raise ValueError(
                f"The ksathorfrac data contains several variables. \
Select the variable to use for ksathorfrac using 'variable' argument."
            )
```

hydromt_wflow/wflow.py
```diff
@@ -1908,6 +1908,61 @@ def setup_soilmaps(
...
+
+        # Create scaled ksathorfrac map
+        dsout = workflows.ksathorfrac(
+            dain,
+            ds_like=self.grid,
+            resampling_method=resampling_method,
+        )
+
+        # Set the grid
+        self.set_grid(dsout)  # TODO of course
```
hboisgon, line 1963:
```suggestion
        self.set_grid(dsout)
```

hydromt_wflow/wflow.py
```diff
@@ -1908,6 +1908,61 @@ def setup_soilmaps(
...
+        self.set_config("input.lateral.subsurface.ksathorfrac", "KsatHorFrac")
```
hboisgon, line 1964: I see you disagree with my suggestion to not have a fixed name for 'KsatHorFrac' but use the variable name in dain instead?

hydromt_wflow/workflows/soilparams.py
```diff
@@ -0,0 +1,53 @@
...
+
+import numpy as np
+import xarray as xr
+
+
+def ksathorfrac(
+    da: xr.DataArray,
+    ds_like: xr.Dataset,
+    resampling_method: str,
+) -> xr.Dataset:
```
hboisgon, line 11: Still think you should return a DataArray

hydromt_wflow/workflows/soilparams.py
```diff
@@ -0,0 +1,53 @@
...
+        dim=da.raster.x_dim, method="nearest", fill_value="extrapolate"
+    )
+
+    # Set outgoing name
+    da.name = "KsatHorFrac"
+    # Set the default no fill value for doubles
+    da = da.fillna(-9999.0)
+    da.raster.set_nodata(-9999.0)
+    # Return as a dataset to be used for 'set_grid'
+    return da.to_dataset()
```
hboisgon, line 53:
```suggestion
    return da
```

tests/test_model_methods.py
```diff
@@ -255,6 +257,37 @@ def test_setup_reservoirs(source, tmpdir, example_wflow_model):
...
+    destination = str(tmpdir.join(model))
+    example_wflow_model.set_root(destination, mode="w")
+
+    # Build the map
+    example_wflow_model.setup_ksathorfrac(
+        ksat_fn=da,
+    )
+    # Write and read the map
+    example_wflow_model.write_grid()
+    example_wflow_model.read_grid()
```
hboisgon, line 281: Don't think you need these lines in your test, do you?

hboisgon (changes requested):
Hi @dalmijn, nice that you separated the ksat part from the the predefined catalogs. When the changes are made in core we can apply then in the plugin.

For this PR, I see the suggestions from my previous reviews were not implemented. Do you disagree with them then?

Also you need to update this page in the docs: docs/user_guide/wflow_model_setup.rst to add setup_ksathorfrac to the setup methods list of wflow.

Else I think almost ready to merge!

hydromt_wflow/workflows/soilparams.py
dalmijn: No It should return a Dataset

tests/test_model_methods.py
dalmijn: Maybe a bit of legacy now, I wasnt able to read the results in Python previously without doing this. I shall check that.

hydromt_wflow/wflow.py
dalmijn: Well, yes. In my opinion it is the ksathorfrac map. This makes is much clearer within the settings file.
We're runnning wflow with only one ksathorfrac map anyway.

hydromt_wflow/wflow.py
dalmijn: Good one.

hydromt_wflow/wflow.py
dalmijn: Agreed!

hydromt_wflow/wflow.py
dalmijn: Agreed!

hydromt_wflow/wflow.py
dalmijn: Allright!

hydromt_wflow/workflows/soilparams.py
hboisgon: Yes a DataArray can be passed and you can even change the name when doing so:
``` self.set_grid(da, name="KsatHorFrac") ```

hydromt_wflow/wflow.py
JoostBuitink: Just to put my thoughts in: in case a user wants to try out different versions of the ksathorfrac maps (either through a different `ksat_fn` or through a different `variable`), I think it is nice these maps will be added to the staticmaps with a recognizable name, rather then always being set to `KsatHorFrac`. Maybe we can use a setup like `khfrac_{variant}`?

hydromt_wflow/workflows/soilparams.py
dalmijn: Allright, then that seems better!

tests/test_model_methods.py
dalmijn: I found my original reason, the data is clipped when written and I wanted to check based on that.

hboisgon (approved):
Hi @dalmijn,

I saw you agreed with my docs and error message suggestions but did not commit or change them. Also in my main review message I asked if you could update also the wflow_model_setup list of components.

I hope you don't mind but to speed up the process I made a commit for this:

- add the suggestions you agreed to
- update the wflow_model_setup docs
- a few extra changes to make clear that ksathorfrac workflow returns a DataArray

The rest of your changes looked good :)

If you are okay with the changes I committed than we can merge this PR.

PR #257, opened 2024-03-21
add setup_pet_forcing method

### LINKED ISSUE #69: Functionality to resample existing PET data
Currently PET is obtained only by computing it using De Bruin or Makking in setup_temp_pet_forcing.
But existing PET data cannot be used directly.

Several implementations:

- in setup_temp_pet_forcing, replace skip pet by a compute_pet flag that if False resample existing pet data and if True compute
- split into two distincts setup_temp_forcing and setup_pet_forcing components
- or three components: setup_temp_forcing, setup_pet_forcing and setup_temp_pet_forcing (combined for temp downscaling)

@DirkEilander do you have any preference? 
I have a feeling the second option might help with memory issue but then you potentially would downscale temperature twice if you compute pet.

hboisgon:
## Issue addressed
Fixes #69

## Explanation
For now just simple reprojection of PET without any complex correction when downscaling. This can be investigated in the future.

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -44,6 +44,7 @@ Setup components
    WflowModel.setup_config_output_timeseries
    WflowModel.setup_precip_forcing
    WflowModel.setup_temp_pet_forcing
+   WflowModel.setup_pet_forcing
    WflowModel.setup_constant_pars
    WflowModel.setup_1dmodel_connection
    WflowModel.setup_grid_from_raster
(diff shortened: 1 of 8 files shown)


hboisgon:
Thanks for the review @dalmijn.
What you mention is not really related to this PR is it?

In ``write_forcing``, ``time_units`` is only used if ``freq_out`` is not None ie if you wan to split your forcing into several files when writing. Because in this case the user already needs to define manually ``freq_out`` then also defining the ``time_units`` should not be that big of an issue I think.
If you write into a single file, xarray automatically uses the first time stamp of the data as the time_units but with several files wflow needs this to be common.

dalmijn:
@hboisgon, I agree. I approve this PR then.

dalmijn (approved):
LGTM.

PR #265, opened 2024-04-11
update ldd example with updates in core

DirkEilander:
## Issue addressed
Improved the LDD example based on new functionality in core.

Requires https://github.com/Deltares/hydromt/pull/305


hboisgon (approved):
LGTM!

Issue #262, opened 2024-04-02
reservoir not detected when not fully contained in 


aweerts:
### HydroMT-Wflow version checks

- [X] I have checked that this issue has not already been reported.
- [X] I have checked that this bug exists on the latest version of HydroMT-Wflow.

### Reproducible Example

Reservoirs ignored even if >99% is in subbasin. This happens for 100+ basins when deriving 3200 models
hack/solution 2022 (Dirk): waarschijnlijk is het voor deze case op te lossen door predicate="contains" weg te laten.
Suggestion to add this as an option in hydromt-wflow 

See:
hydromt build wflow test_5864 -r "{'subbasin': [123.970833333057,42.2791666666327], 'strord': 4,'bounds':[123.908333333333,42.0658333333333,124.9825,42.4583333333333]}" -i wflow_build_1km_vito_hydromt3.ini -d p:\wflow_global\hydromt\deltares_data.yml -vv

![image](https://github.com/Deltares/hydromt_wflow/assets/11685807/8abc1484-111c-43b9-b994-7c8a4bce7681)
![image](https://github.com/Deltares/hydromt_wflow/assets/11685807/9bb975ef-84ee-446c-bdac-00cb4fa194af)

### Current behaviour

reservoir ignored

### Desired behaviour

when option selected reservoir appears

hboisgon:
Maybe try and solve together with #182

atsiokanos:
Indeed this can be solved by changing the `predicate` (see config below) which is a bit hidden in `kwargs`. So I would say this is more of a documentation thing than a bug. Might be good to include it in the docs/method in a more explicit way.

```
  - setup_reservoirs_simple_control:
      reservoirs_fn: hydro_reservoirs  # source for reservoirs shape and attributes
      timeseries_fn: gww           # additional source for reservoir are timeseries to compute reservoirs, Either 'gww' using gwwapi or 'jrc' using hydroengine.
      min_area: 1           # minimum lake area to consider [km2]
      predicate: "intersects"
```

alimeshgi:
Closing this issue. This is already solvable by passing predicate: \"intersects\" to setup_reservoirs_simple_control (currently hidden in kwargs). The broader waterbody delineation improvement is tracked in #182. Documentation should be improved to surface this option.

PR #266, opened 2024-04-24
Real wflow basins

### LINKED ISSUE #173: Saving the actual wflow basins in staticgeoms
So far when building a wflow model we add a **basins.geojson** file in staticgeoms folder. This staticgeoms is however the basin delineation based on the hydrography_fn (merit_hydro) at its resolution and its basin IDs.

I think it would make more sense that **basins.geojson** would be directly the wflow subcatch delineation and IDs as in wflow_subcatch.map and maybe save a second shape for hydrography_fn (merit_hydro) shape and ID, for example **basins_{hydrography_fn}.geojson** or **basins_high_resolution.geojson**.

The current behavior is prone to error, as for example to clip a wflow model, IDs from wflow_subcatch map and not from merit_hydro should be used (the ones in basins.geojson). Also later if that file is used to compute basin statistics, this may lead to small errors.

dalmijn:
## Issue addressed
Fixes #173, #83 

## Explanation
Code.

## Checklist

No.


hydromt_wflow/wflow.py
```diff
@@ -239,25 +255,17 @@ def setup_basemaps(
...
-                raise ValueError(
-                    f"The model resolution {res} should be smaller than 1 degree \
-(111km) for geographic coordinate systems. "
-                    "Make sure you provided res in degree rather than in meters."
-                )
+        # Set name based on scale_factor
+        basins_suffix = ""
+        if scale_ratio != 1:
+            basins_suffix = "_highres"
+        self.set_geoms(geom, name=f"basins{basins_suffix}")
```
hboisgon, line 267: I think you can simplify and actually only set if scale_ratio != 1.
And right after that call self.basins to set it for all cases

hydromt_wflow/wflow.py
```diff
@@ -285,6 +293,8 @@ def setup_basemaps(
         # Rename and add to grid
         rmdict = {k: v for k, v in self._MAPS.items() if k in ds_base.data_vars}
         self.set_grid(ds_base.rename(rmdict))
+        # Call basins once to set it
+        self.basins
```
hboisgon, line 297: move this up

hydromt_wflow/wflow.py
```diff
@@ -1251,7 +1261,7 @@ def setup_gauges(
             kwargs.update(crs=code)
             gdf_gauges = self.data_catalog.get_geodataframe(
                 gauges_fn,
-                geom=self.basins,
+                geom=self.basins_highres,
```
hboisgon, line 1264: For everywhere where you replaced with basins_highres, you should actually replace with basins.
This is a much better match. Imagine you have a gauge at high res that is just out of the basin but is included in the low res. Or for areamap, you may get nodata cell just because your cut is not the exact cut. I think we have buffers by default to in the get_data method to always read a little bit more anyhow but using basins is more consistent.

Something else, not all models will have a basins_highres geoms ie model build with a previous hydromt-wflow version and models built without hydromt. The last ones will always have basins as if basins is not in geoms then it is derived from staticmaps wflow_subcatch.

hydromt_wflow/wflow.py
```diff
@@ -3684,13 +3702,21 @@ def basins(self):
...
         return gdf
 
+    @property
+    def basins_highres(self):
+        """Returns a high resolution basin(s) geometry."""
+        if "basins_highres" in self.geoms:
+            gdf = self.geoms["basins_highres"]
+        else:
+            gdf = self.basins
+        return gdf
```
hboisgon, line 3718: I dont think it's needed :)

docs/changelog.rst
```diff
@@ -21,6 +22,7 @@ Added
 
 Changed
 -------
+- Basins geometry (**basins**) is now based on the actual wflow model resolution basins, instead of based on the supplied DEM
```
hboisgon, line 25: Add PR number

hboisgon (commented):
Hi Brendan, nice! And nice find for the set_precision method, rounding for geodataframe was actually not so easy not too long ago I did not realize...
Just a couple of comments but I dont think we need a really keep track of the basins_highres, it's just nice if it's in your geoms but that's it.

Could you maybe update the examples with the new basins and basins_highres geojson ? I think we are really relaxed in the tests so maybe that's why they did not complain but I think good to keep the examples up to date with this.

hydromt_wflow/wflow.py
dalmijn: Yeah, I was not quite sure which methods were depended on the high res map (as all old functions essentially used that). 
Thanks for the clarification @hboisgon!

docs/changelog.rst
dalmijn: Yup.

hydromt_wflow/wflow.py
dalmijn: That seems fair!

hydromt_wflow/wflow.py
dalmijn: Can't, it is derived from grid when not read from geoms. And the grid with the basins is first set in the line above where I call `self.basins`.

hydromt_wflow/wflow.py
```diff
@@ -3605,7 +3621,7 @@ def read_tables(self, **kwargs):
         if len(fns) > 0:
             for fn in fns:
                 name = basename(fn).split(".")[0]
-                tbl = pd.read_csv(fn)
+                tbl = pd.read_csv(fn, float_precision="round_trip")
```
hboisgon, line 3624: Why this?? And only on the read?
It's not very well documented in pandas but this seems to relate to string precision and not numbers.

hydromt_wflow/wflow.py
```diff
@@ -3684,7 +3700,6 @@ def basins(self):
                 .set_index("value")
                 .sort_index()
             )
-            gdf.index.name = self._MAPS["basins"]
```
hboisgon, line 3687: Why remove this line? This would change the name of the index column from wflow_subcatch to value. Apparently we don't use this in other places in the code but why change?

hydromt_wflow/wflow.py
```diff
@@ -3163,14 +3171,22 @@ def read_staticgeoms(
     def write_geoms(
         self,
         geom_fn: str = "staticgeoms",
+        precision: int = 5,
```
hboisgon, line 3174: I thought we said we would change default precision to 6?

hboisgon: And maybe in the code, you could check is the mod.crs.is_projected and then adjust the default value to 1 for 0.1 meter? And document it in the docstring?
Which I now see should maybe be updated?

hboisgon (changes requested):
Thanks for updating the examples model @dalmijn ! Not sure why the inmaps for some models got updated as well?

Also why did you add two new changes to the code? In read_tables and the index column name in basins?

You are going to hate me but I had more time to think and I actually think you got it right the first time with using basins_highres when available to filter hydrological objects such as lakes/reservoirs/glaciers. So I wonder if we should not put it back in the end for these three? In a way, it's a temporary fix for the Piave case before we address issue #262 properly but as we are talking of releasing, I'm a little uncomfortable to release without the basins high res... What do you think?
For gauges I wonder. For discharge gauges, it may help in very few cases but for precipitation it may make it worse. So maybe here keep low res basins.

hydromt_wflow/wflow.py
dalmijn: It broke the tests

hydromt_wflow/wflow.py
dalmijn: This has more to do with testing (as tests are written for entire dataframes; with writing them to the drive in between). This causes precision issues. The tables test passed last time out of sheer luck. Setting the precision like this does impact performance (when reading) a little bit, but that is not noticable in this case.

dalmijn:
@hboisgon, I do think the low res basins are better for consistency, but I do agree with you that for now (until #262 is fixed) that the highres basins are then better as a temporary fix. I'll change it back. Good thinking.

hydromt_wflow/wflow.py
hboisgon: Which test did it break? Maybe actually using `wflow_subcatch` rather than `value` as name is better no? While we are breaking behavior of self.basins anyhow :)

hboisgon (commented):
Thanks for updating @dalmijn !
Two last remarks:
- the index name of self.geoms["basins"]
- the update of the examples. I agree that the geosjon should change with the new rounding precision but why inmaps. staticmaps and wflow_sbm files should change?

hydromt_wflow/wflow.py
```diff
@@ -3163,14 +3171,25 @@ def read_staticgeoms(
...
     ):
         """Write geoms in <root/geom_fn> in GeoJSON format."""
         # to write use self.geoms[var].to_file()
         if not self._write:
             raise IOError("Model opened in read-only mode")
         if self.geoms:
             self.logger.info("Writing model staticgeom to file.")
+            # Set projection to 1 decimal if projected crs
+            if self.crs.is_projected:
+                precision = 1
```
hboisgon, line 3184: You added this since my last review no? Here the user cannot change that value at all the way you now coded it and for models like dflow fm precision of 10cm may not be good enough. Maybe you could instead set precision to None and then if precision is None and crs is projected then precision is 1 (maybe 2 is better?) and else 6? Then the user is fully in control again.

hboisgon (approved):
Thanks for the adjustments @dalmijn 
I approve already but I saw one last thing that you changed:

- if crs is projected, users cannot change the precision in write_geoms

Can you still update before merging? While you are at it, also add the change in precision of the geoms file to the changelog.

hydromt_wflow/wflow.py
dalmijn: It was after your first review, we had a call concerning the precision when talking about degrees (which need quite some precision) and linear units (metres) which need less. I do agree that having it fixed is not ideal. I'll fiddle around with it a bit to make sure users can set it.

PR #272, opened 2024-05-17
Add support for `alpha_h1` parameter and the `GLCNMO` dataset

JoostBuitink:
## Issue addressed
No issue, but fixes the last bullet in https://github.com/Deltares/hydromt_wflow/pull/226#issuecomment-2098170055

Also adds support for the GLCNMO land use dataset. This dataset is used for (global) models to identify regions where rice/paddy fields are located. Adding support for this dataset ensures more consistent model building.

## Explanation
LULC parameters are based on existing tables. Alpha_h1 values are set to zero for all crops (not-natural vegetation), and set to 1 for all natural vegetation, based on the idea that natural vegetation is adapted to the local conditions and can deal with saturated conditions, while crops are likely not able to handle those conditions.

## Checklist
- [x] Update default models (staticmaps) with new alpha_h1 layer
- [x] Update default models (toml) with new alpha_h1 layer


hydromt_wflow/wflow_sediment.py
```diff
@@ -309,17 +309,17 @@ def setup_lulcmaps(
...
+            "landuse": None,
+            "Cov_River": None,
+            "Kext": "input.vertical.kext",
+            "N": None,
+            "PathFrac": "input.vertical.pathfrac",
+            "Sl": None,
+            "Swood": "input.vertical.storage_wood",
+            "USLE_C": "input.vertical.usleC",
+            "WaterFrac": None,
+        },
```
JoostBuitink, line 322: @hboisgon, is it indeed correct that these parameters (`lateral.land.n`, `vertical.storage_leaf` and `vertical.waterfrac`) are not needed/specified for wflow_sediment?
I now set their keys to None otherwise the test fails, but wanted to make sure this is indeed correct

hboisgon: I checked in the wflow sediment code and for sure ``lateral.land.n``,  ``vertical.waterfrac`` are not used.8 ``vertical.storage_leaf`` should though. For ``Cov_river`` I see it is not used anymore as it was found the equation was not doing too well. Maybe I would like to add it again in the model at some point so we can keep the values in the mapping table but remove it from the list of params to prepare?

hydromt_wflow/data/parameters_data.yml
```diff
@@ -12,6 +12,7 @@ corine_mapping_default:
       Sl: float32
       Swood: float32
       WaterFrac: float32
+      alpha_h1: int16
```
hboisgon, line 15: It is really an integer (always 0 or 1 like a boolean) or can it be a fraction (0,7 or 0.5 etc)? I would guess float after the reprojection to wflow grid anyhow no?

hydromt_wflow/data/parameters_data.yml
```diff
@@ -86,6 +90,29 @@ vito_mapping_default:
     source_version: 1.0
   path: lulc/vito_mapping.csv
 
+glcnmo_mapping_default:
+  data_type: DataFrame
+  driver: csv
+  kwargs:
```
hboisgon, line 96:
```suggestion
  driver_kwargs:
```
kwargs is now deprecated in favor of driver_kwargs in hydromt. Can you update for the rest as well?

hydromt_wflow/wflow.py
```diff
@@ -900,6 +901,8 @@ def setup_lulcmaps(
         * **PathFrac** map: The fraction of compacted or urban area per grid cell [-]
         * **WaterFrac** map: The fraction of open water per grid cell [-]
         * **N** map: Manning Roughness [-]
+        * **alpha_h1** map: Root water uptake reduction at soil water pressure head h1
+          (0.0 or 1.0) [-]
```
hboisgon, line 905: If should be integer can you update here?
```suggestion
          (0 or 1) [-]
```

hydromt_wflow/wflow.py
```diff
@@ -908,12 +911,12 @@ def setup_lulcmaps(
...
             Path to a mapping csv file from landuse in source name to
             parameter values in lulc_vars. If lulc_fn is one of {"globcover", "vito",
-            "corine", "esa_worldcover"}, a default mapping is used and this argument
-            becomes optional.
-        lulc_vars : list
-            List of landuse parameters to keep.
+            "corine", "esa_worldcover", "glmnco"}, a default mapping is used and this
+            argument becomes optional.
+        lulc_vars : dict
+            Dictionary of landuse parameters to keep and their internal wflow name.
```
hboisgon, line 917:
```suggestion
            Dictionary of landuse parameters in ``lulc_mapping_fn`` columns to prepare and their internal wflow name (or None to skip adding to the toml).
```

hydromt_wflow/wflow.py
```diff
@@ -934,12 +937,17 @@ def setup_lulcmaps(
...
             ds_like=self.grid,
             df=df_map,
-            params=lulc_vars,
+            params=list(lulc_vars.keys()),
             logger=self.logger,
         )
         rmdict = {k: v for k, v in self._MAPS.items() if k in ds_lulc_maps.data_vars}
         self.set_grid(ds_lulc_maps.rename(rmdict))
 
+        # Add entries to the config
```
hboisgon, line 946: This is a nice addition and we should do this for all the setup methods but let's so that in another PR. I also wonder if then we should start simplifying and removing from our template config. For now not but let's do this when we implement this for all setup methods

hydromt_wflow/wflow_sediment.py
hboisgon: Can you then also update the docstring similar to the wflow one?

tests/conftest.py
```diff
@@ -25,7 +25,15 @@
 def example_wflow_model():
     logger = logging.getLogger(__name__)
     root = join(EXAMPLEDIR, "wflow_piave_subbasin")
-    mod = WflowModel(root=root, mode="r", data_libs="artifact_data", logger=logger)
+    mod = WflowModel(
+        root=root,
+        mode="r",
+        data_libs=[
+            "artifact_data",
+            "https://github.com/Deltares/hydromt_wflow/releases/download/v0.5.0/wflow_artifacts.yml",
```
hboisgon, line 33: For all three tests, why do we need this? I don't think you test the GLCNMO data so far no?

hboisgon (commented):
Thanks for adding this @JoostBuitink !
Just very small comments before this is ready, I'll let you look into it.
Also nice that you added now the updating of the names in the toml file!

hydromt_wflow/wflow.py
JoostBuitink: Yeah I agree that that would be much nicer and much more robust. Indeed something to fix in another PR, but decided to already make a small step in this PR.

hydromt_wflow/data/parameters_data.yml
JoostBuitink: Fixed it!

tests/conftest.py
JoostBuitink: That was indeed incorrectly copied from the demand branch. Removed it for this PR!

hydromt_wflow/data/parameters_data.yml
hboisgon: Are you sure? I think the default method to reproject is average for all and only mode for landuse. This is set in workflows.landuse.py in a dict at the top of the file. Or does the integer type means that conversion to 0 and 1 will be done after average because of the type (not sure how they convert 0.4, 0.5 and 0.6 with this). Maybe mode method might be better to use then for alpha_1 like for landuse classes?

JoostBuitink: Yeah apparently the datatype prevented it from going to floating point values, but this does indeed feel a bit tricky to rely on (also if users don't specify the dtype correctly). I have added mode resampling for this parameter!

hydromt_wflow/workflows/landuse.py
```diff
@@ -11,7 +11,7 @@
 __all__ = ["landuse", "lai"]
 
 
-RESAMPLING = {"landuse": "nearest", "lai": "average"}
+RESAMPLING = {"landuse": "nearest", "lai": "average", "alpha_h1": "mode"}
 DTYPES = {"landuse": np.int16}
```
hboisgon, line 15: Maybe also here add the dtype in case users "forgot" in their catalog (or did not do a catalog)?
```suggestion
DTYPES = {"landuse": np.int16, "alpha_h1": np.int16}
```

hboisgon (approved):
Good now @JoostBuitink ! Just one small final comment but for the rest it's okay to merge!

hydromt_wflow/workflows/landuse.py
JoostBuitink: Ah thanks for this suggestion, this is indeed a nice option to include!

PR #286, opened 2024-08-16
Small I/O fixes

### LINKED ISSUE #276: Read/write staticgeoms if dir_input is set in toml

### Reproducible Example

try to read/write geoms if dir_input is "input" in toml

### Current behaviour

In the wflow toml, all input can be located in a separate folder using the dir_input argument in the toml.
We take this into account for grid / staticmaps:

https://github.com/Deltares/hydromt_wflow/blob/fc83571742f5dc251cd87b4f103824d205ac7476/hydromt_wflow/wflow.py#L3131-L3142

But not for staticgeoms. For staticgeoms, we assume the geoms folder is in the same path as staticmaps file but we do not take dir_input into account. So then the model geoms are not read.

### Desired behaviour

Allow to read/write geoms if dir_input is set in the toml file.

### Additional context

_No response_

dalmijn:
## Issue addressed
Fixes number of smaller I/O issues
Fixes #276 

## Explanation
Bit and pieces of code.

## Checklist

No.

### DIFF
--- a/examples/linux64/wflow_piave_clip/wflow_sbm.toml
+++ b/examples/linux64/wflow_piave_clip/wflow_sbm.toml
@@ -1,7 +1,7 @@
 casename = "wflow_sbm"
 calendar = "proleptic_gregorian"
-starttime = 2010-02-02T00:00:00
-endtime = 2010-02-10T00:00:00
+starttime = "2010-02-02T00:00:00"
+endtime = "2010-02-10T00:00:00"
 time_units = "days since 1900-01-01 00:00:00"
 timestepsecs = 86400
 dir_output = "run_default"
--- a/examples/wflow_update_forcing.yml
+++ b/examples/wflow_update_forcing.yml
@@ -21,3 +21,5 @@ write_forcing:                # Write forcing and re-write config if forcing fil
 write_grid:             # to make sure that the staticmaps are written as well, can be skipped if update in the same model folder
 
 write_geoms:             # to make sure that the staticgeoms are written as well, can be skipped if update in the same model folder
+
+write_config:           # to make sure the configurations file is written
--- a/examples/wflow_update_forcing.yml
+++ b/examples/wflow_update_forcing.yml
@@ -22,4 +22,4 @@ write_grid:             # to make sure that the staticmaps are written as well, 
 
 write_geoms:             # to make sure that the staticgeoms are written as well, can be skipped if update in the same model folder
 
-write_config:           # to make sure the configurations file is written
+write_config:           # to make sure the configurations file is written with the updated start and endtime and the forcing file path
--- a/hydromt_wflow/wflow.py
+++ b/hydromt_wflow/wflow.py
@@ -3332,9 +3332,8 @@ Run setup_soilmaps first"
         """
         Read static geometries and adds to ``geoms``.
 
-        Assumes that the `geoms_fn` folder is located in the same folder as the static
-        input data (``input.path_static`` and ``dir_input``). If not found, uses assumes
-        they are in <root/geoms_fn>.
+        Assumes that the `geoms_fn` folder is located relative to root and ``dir_input``
+        if defined in the toml. If not found, uses assumes they are in <root/geoms_fn>.
 
         Parameters
         ----------
@@ -3344,22 +3343,13 @@ Run setup_soilmaps first"
         """
         if not self._write:
             self._geoms = dict()  # fresh start in read-only mode
-        # Assumes geoms_fn is located in the same folder as the static input data
-        dir_mod = dirname(
-            self.get_config(
-                "input.path_static", abs_path=True, fallback="staticmaps.nc"
-            )
-        )
         # Check if dir_input is set and add
         if self.get_config("dir_input") is not None:
-            dir_mod = dirname(
-                join(
-                    self.get_config("dir_input", abs_path=True),
-                    self.get_config("input.path_static", fallback="staticmaps.nc"),
-                )
-            )
+            dir_mod = join(self.get_config("dir_input", abs_path=True), geoms_fn)
+        else:
+            dir_mod = join(self.root, geoms_fn)
 
-        fns = glob.glob(join(dir_mod, geoms_fn, "*.geojson"))
+        fns = glob.glob(join(dir_mod, "*.geojson"))
         if len(fns) > 1:
             self.logger.info("Reading model staticgeom files.")
         for fn in fns:
@@ -3375,8 +3365,7 @@ Run setup_soilmaps first"
         """
         Write geoms in GeoJSON format.
 
-        Assumes the geoms_fn folder is located relative to the staticmaps.nc file.
-        Checks the path of ``geoms_fn`` using both ``input.path_static`` and
+        Checks the path of ``geoms_fn`` using both model root and
         ``dir_input``. If not found uses the default path ``staticgeoms`` in the root
         folder.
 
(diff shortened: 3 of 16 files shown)


examples/linux64/wflow_piave_clip/wflow_sbm.toml
hboisgon, line 1: Does it still work without the quotes? Because wflow can work with and without so hydromt also should

examples/wflow_update_forcing.yml
```diff
@@ -21,3 +21,5 @@ write_forcing:                # Write forcing and re-write config if forcing fil
 write_grid:             # to make sure that the staticmaps are written as well, can be skipped if update in the same model folder
 
 write_geoms:             # to make sure that the staticgeoms are written as well, can be skipped if update in the same model folder
+
+write_config:           # to make sure the configurations file is written
```
hboisgon, line 25:
```suggestion
write_config:           # to make sure the configurations file is written with the updated start and endtime and the forcing file path
```

hydromt_wflow/wflow.py
```diff
@@ -3147,37 +3147,48 @@ def setup_cold_states(
             self.set_config(option, states_config[option])
 
     # I/O
-    def read(self):
+    def read(
+        self,
+        config_fn: Union[Path, str] = None,
+        geom_fn: Union[Path, str] = "staticgeoms",
```
hboisgon, line 3153:
```suggestion
        geoms_fn: Union[Path, str] = "staticgeoms",
```

hydromt_wflow/wflow.py
```diff
@@ -3147,37 +3147,48 @@ def setup_cold_states(
             self.set_config(option, states_config[option])
 
     # I/O
-    def read(self):
+    def read(
+        self,
+        config_fn: Union[Path, str] = None,
```
hboisgon, line 3152: Do we need this? You can specify the config_fn in the model init or global options in the hydromt model_build.yml config. If we do not then I think we should have something similar for staticgeoms ie defined during model init and keep the current behavior.

hydromt_wflow/wflow.py
```diff
@@ -3147,37 +3147,48 @@ def setup_cold_states(
...
         self.logger.info("Model read")
 
-    def write(self):
+    def write(
+        self,
+        config_fn: str = None,
+        grid_fn: Union[Path, str] = "staticmaps.nc",
+        geom_fn: Union[Path, str] = "staticgeoms",
+        forcing_fn: Union[Path, str] = None,
+    ):
```
hboisgon, line 3171: I think if you want to change any default argument of write, you should then call the single write methods one by one. The general write is only meant to be a shortcut for a quick write.

hydromt_wflow/wflow.py
```diff
@@ -3147,37 +3147,48 @@ def setup_cold_states(
...
         """Write the complete model schematization and configuration to file."""
         self.logger.info(f"Write model data to {self.root}")
         # if in r, r+ mode, only write updated components
         if not self._write:
             self.logger.warning("Cannot write in read-only mode")
             return
         self.write_data_catalog()
-        if self.config:  # try to read default if not yet set
```
hboisgon, line 3169: Most filenames are read from the config and as the comment says, we did find some cases where at that stage config was not yet read. To be sure, I would then leave a call to self.config to be sure it is read. You can keep the writing as last.

hydromt_wflow/wflow.py
```diff
@@ -3297,17 +3308,12 @@ def read_geoms(
...
+        Assumes that the `geom_fn` folder is located in the same folder as the root or
+        the folder where the config file is located (which is the root).
         """
         if not self._write:
             self._geoms = dict()  # fresh start in read-only mode
-        dir_default = join(self.root, "staticmaps.nc")
-        dir_mod = dirname(
-            self.get_config("input.path_static", abs_path=True, fallback=dir_default)
-        )
-        fns = glob.glob(join(dir_mod, geom_fn, "*.geojson"))
```
hboisgon, line 3310: I think I did that in case staticmaps path was for example ../../staticmaps. Then the user does not have to think too much about the staticgeoms path as you can assume it will most probably be in the same folder as staticmaps. The one thing I did not take into account and that leads to errors is that in the toml you can now specify a dir_input like you have a dir_output. I think it may be nicer to keep the current behavior and to also handle the dir_input for staticgeoms see #276 
What do you think? 
Else this means the user needs to manually specify the staticgeoms folder and to avoid making it too complex (ie can easily call read/write), I would then put the geom_fn as a __init__ argument of model.
So I would be okay it either 1 we keep the current code and add dir_input or 2) we keep your implementation but geom_fn can be defined in init.

hydromt_wflow/wflow.py
```diff
@@ -3533,9 +3537,9 @@ def write_forcing(
                     f"Not all dates found in precip_fn changing starttime to \
 {start} and endtime to {end} in the toml."
                 )
-                self.set_config("starttime", start.to_pydatetime())
-                self.set_config("endtime", end.to_pydatetime())
-                self.write_config()
+                # Set the strings first
+                self.set_config("starttime", start.strftime("%Y-%m-%dT%H:%M:%S"))
+                self.set_config("endtime", end.strftime("%Y-%m-%dT%H:%M:%S"))
```
hboisgon, line 3542: Why did you change to strings? Wflow can handle both or did you have some errors?

hboisgon (commented):
Nice that you try to tackle some of the remaining I/O issues! We discussed before but I agree to remove write_config from write_forcing and to write it last in model.write.
For the strings of starttime and endtime, I wonder why did you change? As long as both strings or times can be read by hydromt I do not mind so much.
For the staticgeoms, I suggest two solutions that I would be fine with.
For read and write I would keep the current behavior.

hydromt_wflow/wflow.py
dalmijn: More uniform handling of the date typing. It now ran from string to datetime, sometimes back.

hydromt_wflow/wflow.py
dalmijn: I know, but I noticed while using it myself that sometimes I want to write everything but change one name of the output. I made it kwargs only to support the quick `model.write()`. I'd say, it's just a little bit of extra functionality.

hydromt_wflow/wflow.py
dalmijn: Ill change it!

examples/linux64/wflow_piave_clip/wflow_sbm.toml
dalmijn: Hydromt works, but the tests will fail, so I put quotes around them.

examples/wflow_update_forcing.yml
dalmijn: Agreed!

hydromt_wflow/wflow.py
dalmijn: Well most of the time the staticmaps is in the same folder as the config file, i.e. the root directory, so I felt like that the addition of the staticmaps is a bit too much. Therefore, It made, at least to me, not a lot of sense to rely on the staticmaps. Maybe it's best to have a quick call regarding this.

hboisgon:
I directly changed the PR @dalmijn, hope it's okay.
I added some extra things:

- We did not decide but I decide to support the dir_input for staticgeoms. I know already several users that use that option to avoid having too many toml files in their root folder especially for calibration so I think this would be the more robust method.
- For write, I kept your new functionality that filenames can be edited there, I added the states one for completion.
- Added changelog

All the rest was good, thanks for the changes! If you agree with my last modifications, then we can merge :)

Note, I edited the test on ksathorfrac as nowadays read/write netdcf files can be unstable and cause test to hang. The values are different but this is because as we do not write, the catchment mask is not yet applied.

hboisgon (approved):
See the few changes from my commit but the rest looks good thanks!

hydromt_wflow/wflow.py
dalmijn: Agreed!

dalmijn:
@hboisgon, Ill go over it one last time, but it looks good!
Thanks for adding to the PR!

PR #291, opened 2024-09-13
add a test to build model using direct conda forge install

### LINKED ISSUE #287: Error in setup_soilmaps c

### Reproducible Example

Just building a model with setup_soilmaps should create the issue

### Current behaviour

image

### Desired behaviour

Seems to be because of xarray new version v2024.07.0

### Additional context

_No response_

hboisgon:
## Issue addressed
Fixes #287

## Explanation

I investigated why the error did not show up in our tests. There is two reasons for it:

- installing hydromt_wflow from conda-forge results in different packages being installed than installing from an environment file even if all the dependencies listed are the same and version controls are the same. For example, installing from conda-forge gives hydromt 0.9.4 and xarray 2024.9.0 while installing from the environment file gives hydromt 0.10.0 and xarray 2024.3.0
- in our tests, we also build a model with too coarse resolution for this specific error to show. If we use the settings of the piave as in the jupyter notebooks (which is the first thing external user try) rather than the test, then the error showed up. 

So I decided to create a new test workflow where we can check if the easy conda forge install of hydromt-wflow is still able to build models. It will allow us to flag earlier broken conda-foge release but unfortunately if it broken, the only to get the test green again will be to do a new release of hydromt-wflow on conda-forge.

In the meantime, I put a max version on xarray (and numpy) in the pyproject toml and will do the same in the conda forge release. Problem is, I wont be able to test before we actually release on conda forge.

## Checklist

### DIFF
--- a/.github/workflows/test_conda_forge.yml
+++ b/.github/workflows/test_conda_forge.yml
@@ -1,10 +1,11 @@
 name: Test conda forge release
 
 on:
-  push:
-    branches: [ main ]
-  pull_request:
-    branches: [ main ]
+  schedule:
+  # At 00:00 UTC every Monday
+    - cron: "0 0 * * 1"
+  # on demand
+  workflow_dispatch:
 
 jobs:
   Test:
--- a/.github/workflows/test_conda_forge.yml
+++ b/.github/workflows/test_conda_forge.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Install test dependencies
         run: |
-          mamba install -c conda-forge pytest pytest-cov pytest-mock
+          mamba install -c conda-forge pytest pytest-cov pytest-mock pytest-timeout
           pip install gwwapi
 
       - name: Conda info
@@ -52,4 +52,4 @@ jobs:
          conda list
 
       - name: Run model build Tests
-        run: python -m pytest --verbose tests/test_model_class.py::test_build_model
+        run: python -m pytest --verbose tests/test_model_class.py::test_model_build
--- a/.github/workflows/test_conda_forge.yml
+++ b/.github/workflows/test_conda_forge.yml
@@ -1,0 +1,55 @@
+name: Test conda forge release
+
+on:
+  push:
+    branches: [ main ]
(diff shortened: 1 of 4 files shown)


Huite:
> installing hydromt_wflow from conda-forge results in different packages being installed than installing from an environment file even if all the dependencies listed are the same and version controls are the same.

FWIW, if you use pixi the lockfile will guarantee that you get the versions as described.

savente93:
there are a few considerations here: 
- I'm not sure how Conda does it's solving but I do know it is not stable and it has been known to give different versions between runs, so perhaps that is playing into the problem here. 
- As Huite mentions, if these versions are very important I'd consider using something that has a lockfile so you know you can always install the exact dependencies needed (such as pixi. There is also `conda-lockfile` though I don't have experience with that so I'm not sure how well it works) 
- Generally I'm not in favor of "unpassable" tests, though on the other hand, if the current release is broken, that might warrant immediate action. If that is the case I think it's also fine to have it like this, since then it should always be green, though I'm not sure if it has that big of a priority for you. 
- Given that this is about dependencies that perhaps you don't have influence over, (such as new breaking dependencies being released) you might want to run this on a schedule rather than it being attached to PRs since problems can arrise independent of what you do with your code, this would also solve the red PR problem. If you need some pointers on how to run workflows on schedules you can see an example here (https://github.com/Deltares/hydromt/blob/main/.github/workflows/pixi_auto_update.yml) and some explanation here https://www.baeldung.com/cron-expressions

All that being said, the PRs are just a workflow to aid you in development, so as long as you make clear (and hopefully consistent) decisions and communicate them you can do whatever works for you :)

hboisgon:
Thanks for your suggestions @savente93 ! It was very useful :)
I decided to go for a weekly scheduled workflow for the conda test. I think it should be enough and should allow us to spot dependencies or install issues even if we do not work on PRs often.
We also discussed that pixi lock file would only help out pixi users and not conda so for now we test on conda and we can open in our issues to also allow install with pixi for pixi users but will work on that later.

PR #300, opened 2024-10-13
Fix PTF for estimation of residual water content

### LINKED ISSUE #285: Wrong equation for residual moisture content seems to be used

### Reproducible Example

-

### Current behaviour

The function in hydromt-wflow https://github.com/Deltares/hydromt_wflow/blob/29e6c685044d3c67d7e10b1038a625bad0e76b21/hydromt_wflow/workflows/ptf.py#L173 states that the residual moisture content is estimated, but the used equation is actually for the wilting point. See S1 of https://bsssjournals.onlinelibrary.wiley.com/doi/full/10.1111/ejss.12192.

image

- 

### Desired behaviour

-

### Additional context

_No response_

verseve:
## Issue addressed
Fixes #285 

## Explanation
The PTF used from Tóth et al. (2015) estimates the water content at wilting point which can result in an overestimation of the residual water content. This PTF has been replaced by a PTF for residual water content from Rawls and Brakensiek (1989).

## Checklist


verseve:
Tests fail, probably caused by this [issue](https://github.com/Deltares/hydromt/issues/1087).

shartgring (commented):
Hi Willem,

I checked the paper and indeed the equation form Toth refers to the wilting point. After some thought, I do think that it might be worth leaving the PTF of Toth in the code and rename it to thetawp_toth to indicate that it refers to the wilting point. 

Similar to ksatver, we can give the user the option to select the corresponding PTF for thetar in setup_soilmaps. Where the default option will points to thetar_rawls_brakensiek. And include in the docs that using the wilting point for theta_r only typically works for sandy soils.

It might not be used (neither recommended) but I do think it is nice to prevent breaking changes and also improve transparancy by showing the user that this PTF is involved in an explicit way. This may also apply to thetas now that I am inspecting the code a bit more

So summarized I would suggest to leave the PTF of Toth in, make Rawls & Brakensiek the default and include a ptf_thetar (and ptf_thetas?) parameter in setup_soilmaps. What do you think? I could also do this myself as I am nos suggesting a bit more than some small changes to the code

shartgring:
Also saw this chapter in a book: https://www.researchgate.net/file.PostFileLoader.html?id=588ae3ceed99e19c704b6243&assetKey=AS%3A455022786945024%401485497294074

Where they state that:

"The wilting point is an excellent approximation of the residual water content of very coarsetextured (sandy) soils. However, the wilting point at pF ¼ 4.2 likely over-estimates the residual water content of fine-textured (clay) soils where still lots of bound water may be present that collectively (because of the large surface area) may still yield a reasonable liquid flow rate." (pg. 159)

verseve:
> Hi Willem,
> 
> I checked the paper and indeed the equation form Toth refers to the wilting point. After some thought, I do think that it might be worth leaving the PTF of Toth in the code and rename it to thetawp_toth to indicate that it refers to the wilting point.

Hi @shartgring , "to keep it simple" I think it is best to remove this function from the code. I agree, for very coarse textured soils the wilting point is a good approximation for the residual water content. However, I think it will be difficult to decide (for a user) when to use this function: it could be based on the sand percentage, but this check is then required for multiple soil layers (e.g. SoilGrids provides data for 7 soil depths), and for one location you may need then both PTFs depending on the sand percentage.

shartgring (approved):
Thanks for your reply Willem. I think you made a valid point so I'll approve the review

PR #315, opened 2025-01-06
Interpolation of point forcing data (station-based) using wradlib

### LINKED ISSUE #312: Use time series to create forcing files
### Kind of request

Possible enhancement of setup_precip_forcing, or alternatively a new feature. I am working on this in a branch as part of making things easier for stress-testing and model coupling

### Enhancement Description

Currently, only gridded data can be used to create forcing files. In some cases it can be convenient to create forcing files from timeseries (csv), either uniformly or using interpolation methods

### Use case

- When gridded data does not have the desired coverage or temporal resolution, which can be supplied by multiple local stations
- When working with uniform design rainfall events (e.g. to compute boundary conditions, stress-testing, climate analysis)

### Additional Context

Might this be something that would also fit in HydroMT core?

Would this be something to use as an enhancement of the current setup_precip_forcing (and let it detect if precip_fn refers to gridded data or timeseries), or add it as a new method?

And, which interpolations would be good to include? Thiessen polygons, Inverse Distance Weighting, ... ? Literature might be good source for methods, see for example https://cdn.knmi.nl/knmi/pdf/bibliotheek/knmipubIR/IR2009-04.pdf

shartgring:
## Issue addressed
Fixes #312

## Explanation
Consists of a part in wflow.py for reading and checking the forcing data and station data, and a new workflow in workflow/forcing.py where interpolation takes place.

## Checklist
- [X] Updated tests or added new tests
- [X] Branch is up to date with `main`
- [X] Tests & pre-commit hooks pass
- [X] Updated documentation if needed

## Additional Notes

Interpolation types not working in tests
- natural neighbor gives error messages and does not work with N = 3 (which was my initial test, it is now N = 8). Not sure why though
- Cressman interpolation returns a uniform field, so that one needs some extra checking, maybe by directly calling the MetPy function with some test data and seeing what the output looks like. I have only checked the test data which was also processed by hydromt.workflows.precip (the forcing map itself), so maybe the raw data coming out of workflows.forcing.spatial_interpolation needs checking.

01-04-2025: replaced metpy with wradlib


shartgring:
The images from the visual inspection of the test data:

hydromt_wflow_interp_barnes
hydromt_wflow_interp_linear
hydromt_wflow_interp_nearest
hydromt_wflow_interp_rbf

shartgring:
Allright! After some work the code looks good. Only thing to improve is how the gauge stations are loaded. That code is now copied from setup_gauges, but it can be cleaner and the implementation of the GeoDataset can be improved (linking it to precip_fn and not the stations).

The idea is to leave this PR as a draft as MetPy functionality is not working for natural neighbor and Cressman. It would make sense to switch to https://docs.wradlib.org/en/latest/. However, this will not work with the xarray requirement, and to get rid off the pin we need to upgrade to HydroMT version 1.0. Therefore, the PR will be a draft until we know more about the update to version 1.0.

shartgring:
The loading of the stations is now improved (the copy from setup_gauges contained some code based on older HydroMT versions, we can now use new functionality). Also included an option for uniform precipitation

shartgring:
I have included an option to filter stations using the basin geometry and a buffer. This might make it easier to work with large datasets. Would it maybe be logical to include somewhere which stations have been used for interpolation? Ar a possible geom / geojson for visualization? I am doubting as they are not part of the model. Listing all the stations in the log is an alternative, but might end up with a long list

JoostBuitink:
> I have included an option to filter stations using the basin geometry and a buffer. This might make it easier to work with large datasets. Would it maybe be logical to include somewhere which stations have been used for interpolation? Ar a possible geom / geojson for visualization? I am doubting as they are not part of the model. Listing all the stations in the log is an alternative, but might end up with a long list

Great work! I think it would be very nice to have the meteo stations added as a geojson as well! They are not directly part of the model, but can play quite a big role indirectly, so I can imagine that quickly plotting these locations is very valuable.

hydromt_wflow/workflows/forcing.py
```diff
@@ -73,3 +99,170 @@ def pet(
...
+
+    # Determine the dimensions of the resulting interpolated grid
+    x = stations.geometry.x
+    y = stations.geometry.y
+    boundary_coords = get_boundary_coords(x, y)
+    grid_x, grid_y = generate_grid(hres, boundary_coords)
+    
+    # Create empty DataArray to store the interpolated values
+    coords = {"time": forcing.index, "x": grid_x[0, :], "y": grid_y[:, 0]}
+    da_forcing = xr.DataArray(coords=coords, dims=["time", "y", "x"])
```
shartgring, line 217: I am now myself running into memory issues. It might be worth checking if the dimensions of grid_x and grid_y are larger than the model domain and if that is the case, crop da_forcing to the model domain. I expect that in that case `if img.shape != grid_x.shape` will probably be True all of the time, so that one can then be removed

docs/changelog.rst
```diff
@@ -12,6 +12,7 @@ Unreleased
 Added
 -----
 - **setup_ksatver_vegetation**: method to calculate KsatVer_vegetation to account for biologically-enhanced soil structure in KsatVer.
+- **setup_precip_from_point_timeseries**: method to interpolate rainfall station data as model forcing.
```
hboisgon, line 15: Nice! Before it's finished, would be good to update the table with the available methods in the docs as well as the Model and workflows API docs.

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
             precip_out.attrs.update({"precip_clim_fn": precip_clim_fn})
         self.set_forcing(precip_out.where(mask), name="precip")
 
+    def setup_precip_from_point_timeseries(
+        self,
+        precip_fn: Union[str, pd.Series, pd.DataFrame, xr.Dataset],
```
hboisgon, line 2721:
```suggestion
        precip_fn: Union[str, pd.DataFrame, xr.Dataset],
```
The code is complex enough as is so I do not think we need to support pd.Series. I think the get_dataframe method retunrs a dataframe directly and if people want ot pass a python object directly, they can convert it themselves first to a dataframe. If you do not agree let me know!

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+        interp_type: str = "nearest",
+        precip_stations_fn: Union[str, gpd.GeoDataFrame] = None,
+        index_col: str = None,
+        buffer: int = 100,
+        metpy_kwargs: dict = None,
+        **kwargs,
```
hboisgon, line 2727: kwargs are actually not so relevant for the hydromt ``precip`` function so we can simplify into one kwargs.

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+        endtime = self.get_config("endtime")
+        freq = pd.to_timedelta(self.get_config("timestepsecs"), unit="s")
+        mask = self.grid[self._MAPS["basins"]].values > 0
+
+        # Use buffer around basins if provided, else limit to stations covered by basin
+        geom_buffered = (
+            self.basins.buffer(buffer * max(np.abs(self.res)))
+            if buffer
+            else self.basins
+        )
```
hboisgon, line 2806: you can use a buffer directly in the get_data method of the data catalog :)

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+            # Use station ids from gdf_stations when reading the DataFrame
+            station_ids = (
+                gdf_stations[index_col] if index_col else gdf_stations.index
+            )
+
+            gdf_inside = gdf_stations.to_crs(self.crs).within(self.basins.unary_union)
+            self.logger.info(
+                f"""Found {len(station_ids)} stations in {precip_stations_fn} \
+(of which {gdf_inside.sum()} are located inside the basin), using a buffer \
+of {buffer} cells."""
```
hboisgon, line 2838: This info can be moved to interp function. This will avoid repeat if we want to implement for temperature/pet or when we do support geodataset.

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+
+        # Transform station coordinates to model crs
+        gdf_stations = gdf_stations.to_crs(self.crs)
+
+        # Use model resolution for the interpolated grid
+        if not np.allclose((np.abs(self.res)), 1e-6):
+            self.logger.info(
+                f"""Mismatch in horizontal and vertical resolution ({self.res}), \
+using finest resolution during interpolation by MetPy."""
+            )
```
hboisgon, line 2919: This info can be moved to interp function. This will avoid repeat if we want to implement for temperature/pet or when we do support geodataset.

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+        precip = precip.raster.reproject_like(self.grid[self._MAPS["elevtn"]])
+        if box(*self.bounds).contains(box(*precip.raster.bounds)):
+            # Try to stick to original interpolation method
+            fill_method = (
+                    interp_type if (interp_type in ["linear", "nearest", "cubic"])
+                    else "rio_idw"
+                )
+            precip = precip.raster.interpolate_na(
+                method=fill_method, extrapolate=True
+            )
```
hboisgon, line 2962: This info can be moved to interp function. This will avoid repeat if we want to implement for temperature/pet. It also keeps the alreday long wflow.py file a little more lean.

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+                f"""Mismatch in horizontal and vertical resolution ({self.res}), \
+using finest resolution during interpolation by MetPy."""
+            )
+        self.logger.info(f"""Starting interpolation of data: {df_precip.shape[1]} \
+stations and {df_precip.shape[0]} timesteps.""")
+        
+        # Calling interpolation workflow
+        precip = workflows.forcing.spatial_interpolation(
+            forcing=df_precip,
+            stations=gdf_stations,
```
hboisgon, line 2926: You can use a geodataset object here instead of df+stations

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+            interp_type=interp_type,
+            hres=min(np.abs(self.res)),
```
hboisgon, line 2928: pass self.grid instead to do more things in the function

hydromt_wflow/workflows/forcing.py
```diff
@@ -73,3 +99,170 @@ def pet(
...
+    if np.isnan(forcing.values).any():
+        logger.warning(
+            """Forcing data contains NaN values. \
+These will be skipped during interpolation.""")
+
+    if interp_type not in interpolation_supported.keys():
+        raise ValueError(f"Interpolation type {interp_type} not recognized.")
+
+    elif interpolation_supported[interp_type]:
+        # Create a dictionary of the arguments to show in logging
```
hboisgon, line 185: Nice!

hydromt_wflow/workflows/forcing.py
hboisgon: Not sure I fully follow but let's see if the memory issues happen a lot or not also if you prepare focing for less timesteps. I think bacause metpy using numpy and not xarray/dask there is not so much we can do to avoid memory errors anyhow?

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,267 @@ def setup_precip_forcing(
...
+        else:
+            raise ValueError(f"Data source {precip_fn} not recognized.")
+
+        df_precip = df_precip.astype("float32")
+
+        # Check number of stations, at least 3 needed to interpolate spatially
+        if (len(df_precip.columns) < 3) & (interp_type != "nearest"):
+            raise ValueError(
+                f"""Only {len(df_precip.columns)} found with timeseries, which is \
+not sufficient for interpolation using {interp_type} (3 or more required)."""
```
hboisgon, line 2908: Maybe it's more my personal preference but I do not like the line return so much... especially if there is a lot of if else, it visually breaks the indent level of the code. And keeping the indent usually does not add more lines to the code anyhow :)

hydromt_wflow/workflows/forcing.py
```diff
@@ -73,3 +99,170 @@ def pet(
...
+    coords = {"time": forcing.index, "x": grid_x[0, :], "y": grid_y[:, 0]}
+    da_forcing = xr.DataArray(coords=coords, dims=["time", "y", "x"])
+    da_forcing.raster.set_crs(stations.crs)
+    
+    empty_timesteps = []
+    last_logged_progress = 0
+    for i, (timestep, observations) in enumerate(forcing.iterrows()):
+        
+        # TODO is there a nicer way using the logger?
+        progress = ((i + 1)/len(forcing))*100
```
hboisgon, line 225: I do not mind so much, this is actually quite elegant :) else we can install the additional tqdm package but this means more dependencies once again so not sure how worth it it would be.

hboisgon (commented):
Very nice work @shartgring ! I tested the rbf method for my basin and works like a charm :)
I did some modifications to make it work so I already implemented most of my review comments, hope you don't mind... 
I still left comments of the things I changed before comitting so that you can more easily keep track of what I did.

hydromt_wflow/wflow.py
```diff
@@ -2715,6 +2716,177 @@ def setup_precip_forcing(
             precip_out.attrs.update({"precip_clim_fn": precip_clim_fn})
         self.set_forcing(precip_out.where(mask), name="precip")
 
+    def setup_precip_from_point_timeseries(
+        self,
+        precip_fn: Union[str, pd.DataFrame, xr.Dataset],
+        interp_type: str = "nearest",
+        precip_stations_fn: Optional[Union[str, gpd.GeoDataFrame]] = None,
+        index_col: Optional[str] = None,
+        buffer: int = 100,
```
hboisgon, line 2725: If we keep to the hydromt buffer, the unit is actually in meters... so good to document it well or maybe change the unit here in km?

shartgring: I think that the buffer here was actually in cells! I expect that this might change as we implement the functionalities from wradlib so we can leave it as is now

hydromt_wflow/wflow.py
shartgring: Allright! Maybe I thought that get_dataframe would return a series when only 1 stations was included :)

shartgring:
Replaced metpy with wradlib now that newer xarray versions are supported :)

Tests are passing and all interpolation methods work fine so looks like it is ready for review

hydromt_wflow/wflow.py
```diff
@@ -2868,6 +2869,181 @@ def setup_precip_forcing(
...
+
+            # Convert to geodataset
+            da_precip = hydromt.vector.GeoDataArray.from_gdf(
+                gdf=gdf_stations,
+                data=df_precip,
+                name="precip",
+                index_dim=None,
+                dims=["time", gdf_stations.index.name],
+                keep_cols=False,
+                merge_index="gdf",
```
hboisgon, line 3016: This ensure that only index in gdf_stations will be kept. So the lines above are not needed :)

```
# Filter df_precip by stations within the buffered basin
df_precip = df_precip[gdf_stations.index]
```

hydromt_wflow/workflows/forcing.py
```diff
@@ -73,3 +92,155 @@ def pet(
...
+    forcing = forcing.vector.to_crs(crs).astype("float32")
+
+    # Extract station coordinates
+    gdf_stations = forcing.vector.to_gdf()
+    src = np.vstack((gdf_stations.geometry.x, gdf_stations.geometry.y)).T
+
+    # Some info/checks on the station data
+    nb_stations = len(gdf_stations)
+    basins = ds_like["wflow_subcatch"].raster.vectorize()
+    nb_inside = gdf_stations.within(basins.unary_union).shape[0]
```
hboisgon, line 170: This is a little wflow specific maybe? Fine for now of course but if we want others to reuse you could provide a mask_name = None argument (for wflow then it will be self._MAPS["basins"]) or add an if "wflow_subcatch" in ds_like.

tests/test_model_methods.py
```diff
@@ -815,6 +815,64 @@ def test_setup_floodplains_2d(elevtn_map, example_wflow_model, floodplain1d_test
     )
 
 
+def test_setup_precip_from_point_timeseries(
```
hboisgon, line 818: Super nice! Did you lso check for geodataset if it works?
You can add a test by first creating a geodataset using df_precip_stations and gdf_precip_stations and then calling the method :)

hboisgon (changes requested):
Very nice improvements @shartgring ! This will definitely become a useful function I think.
I have a few comments mostly on documentation.
And to keep in with the docs, you should not forget to add this function to the hydromt_wflow docs itself :) There are two places for that: api.rst (both for the wflow function and the new forcing workflow) and then update the table for all setup methods for wflow in user_guide/wflow_model_setup.rst

hydromt_wflow/wflow.py
```diff
@@ -2868,6 +2869,181 @@ def setup_precip_forcing(
...
+        * externaldriftkriging: Kriging interpolation including an external drift, \
+        see wradlib documentation for a full explanation: \
+        `wradlib.ipol.ExternalDriftKriging <https://docs.wradlib.org/en/latest/generated/wradlib.ipol.ExternalDriftKriging.html>`.
+
+
+        Parameters
+        ----------
+        precip_fn : str, pd.DataFrame, xr.Dataset
+            Precipitation source as DataFrame or GeoDataset, see
+            data/forcing_sources.yml. The first column should contain time and the other
```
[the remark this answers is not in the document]
shartgring: Indeed! I also removed it in `setup_precip_forcing` and `setup_temp_pet_forcing`

hydromt_wflow/workflows/forcing.py
shartgring: Good point! I have included it

tests/test_model_methods.py
shartgring: Yes done it and test is passing 👍

hboisgon (approved):
Great thanks a lot for the updates! It's good to go now :)

PR #320, opened 2025-01-20
add setup_lulcmaps_from_vector

hboisgon:
## Issue addressed
No issue was created.

## Explanation
Often local landuse data is in vector/shapefile format rather than vector. This function allows to use vector by first rasterizing to a high resolution grid (resolution from user) before calling the classic ``setup_lulcmaps`` function.

An equivalent ``setup_lulcmaps_from_vector_with_paddy`` was not added as users will be able to save the rasterized version of the landuse in the ``root/maps`` folder, so they can first call ``setup_lulcmaps_from vector`` and then ``setup_lulcmaps_with_paddy`` using the previously saved rasterized version of the landuse.

## Checklist


hydromt_wflow/workflows/landuse.py
```diff
@@ -83,6 +85,82 @@ def reclass(x):
...
+    logger=logger,
+):
+    """
+    Derive several wflow maps based on vector landuse-landcover (LULC) data.
+
+    The vector lulc data is first rasterized to a raster map at the model resolution
+    or at a higher resolution specified in ``lulc_res``.
+
+    Lookup table `df` columns are converted to lulc classes model
+    parameters based on literature. The data is remapped at its original resolution
```
shartgring, line 105: at its original resolution -> at its rasterized resolution (lulc_res)

hydromt_wflow/wflow.py
```diff
@@ -947,6 +947,131 @@ def setup_lulcmaps(
...
+        )
+        # read landuse map
+        gdf = self.data_catalog.get_geodataframe(
+            lulc_fn,
+            geom=self.region,
+            buffer=2,
+            variables=["landuse"],
+        )
+        if save_raster_lulc:
+            lulc_out = join(self.root, "maps", "landuse_raster.tif")
```
shartgring, line 1052: The API mentions that an attribute WflowModel.maps exists, but I have trouble finding it. Or is it the same as WflowModel._MAPS? Reason why I ask is that if we export maps, it might be nice to include them during clipping or updating a model (similar to the tables and geoms).

What do you think?

hydromt_wflow/wflow.py
```diff
@@ -947,6 +947,131 @@ def setup_lulcmaps(
...
+            Resolution of the intermediate rasterized landuse map. The unit (meter or
+            degree) depends on the CRS of lulc_fn (projected or not). By default None,
+            which uses the model resolution.
+        all_touched : bool, optional
+            If True, all pixels touched by the vector will be burned in the raster,
+            by default False.
+        save_raster_lulc : bool, optional
+            If True, the high resolution rasterized landuse map will be saved to
+            maps/landuse_raster.tif, by default False.
+        """
```
shartgring, line 1035: See Also
--------
workflows.landuse_from_vector

hydromt_wflow/workflows/landuse.py
```diff
@@ -83,6 +85,82 @@ def reclass(x):
...
+        Logger object.
+
+    Returns
+    -------
+    ds_out : xarray.Dataset
+        Dataset containing gridded landuse based maps
+    """
+    # rasterize the vector data
+    logger.info("Rasterizing landuse map")
+    if lulc_res is None:
```
shartgring, line 135:
```
    if lulc_res is None:
        grid_like = ds_like
    else:
        grid_like = create_grid_from_bbox(
            gdf.total_bounds,
            res=lulc_res,
            crs=gdf.crs,
            align=True,
        )
```

I notice that many workflows seem to have as little optional arguments as possible. So to do the same to this workflow, the block above might be something to move to the setup method and only include `ds_like` as argument in the workflow

shartgring: Or maybe not worth the hassle?

hydromt_wflow/workflows/landuse.py
```diff
@@ -83,6 +85,82 @@ def reclass(x):
...
+    logger : logging.Logger, optional
+        Logger object.
+
+    Returns
+    -------
+    ds_out : xarray.Dataset
+        Dataset containing gridded landuse based maps
+    """
+    # rasterize the vector data
+    logger.info("Rasterizing landuse map")
```
shartgring, line 134:
```
    if lulc_res is None:
        grid_like = ds_like
    else:
        grid_like = create_grid_from_bbox(
            gdf.total_bounds,
            res=lulc_res,
            crs=gdf.crs,
            align=True,
        )
```

I think that the bounds of the saved map will change when lulc_res is provided or not. As in, grid_like has the bounds of the WflowModel, and gdf has its own bounds. So when lulc_res = None, the bounds may be different and not only the resolution

This would normally be no problem as the gdf gets clipped inside the setup method of the WflwoModel, but when using the workflow on its own (which will probably happen as it is very useful :) ) it might be inconsistent. Maybe can be solved by using ds_like.raster.box.to_crs(gdf.crs).total_bounds or something similar? That would ensure that using lulc_res only changes the resolution and not the bounds

hydromt_wflow/workflows/landuse.py
```diff
@@ -364,3 +442,63 @@ def add_paddy_to_landuse(
...
+    -------
+    usle_c : xr.DataArray
+        Updated USLE C map.
+    """
+    # Add a USLE_C column with default value
+    logger.info(
+        "Correcting USLE_C with planted forest and orchards"
+        "using {planted_forest_fn}."
+    )
+    if planted_forest is None:
```
shartgring, line 482:
```
    if planted_forest is None:
        logger.warning("No Planted forest data found within domain.")
        return None
```
This might be something to shift to the setup method and make planted_forest a required argument for the workflow

hydromt_wflow/workflows/landuse.py
```diff
@@ -364,3 +442,63 @@ def add_paddy_to_landuse(
...
+
+    Parameters
+    ----------
+    planted_forest : geopandas.GeoDataFrame
+        GeoDataFrame containing planted forest data. Required columns are: 'geometry',
+        and optionally 'forest_type' to find orchards.
+    ds_like : xr.Dataset
+        Dataset at model resolution. Required variables are 'USLE_C'.
+    planted_forest_c : float, optional
+        USLE C value for planted forest, by default 0.0881.
```
shartgring, line 466: Do you have a link or reference or DOI for the 0.0881 or 0.2188 values to include in the docs? To prevent that people might ask in the future where these 'magical values' come from ;)

shartgring: Ah I see now that they were already mentioned in the setup method. Still nice to consider adding some kind of reference

hydromt_wflow/wflow_sediment.py
```diff
@@ -388,29 +379,134 @@ def setup_lulcmaps(
...
+                self.grid,
+                planted_forest_c=planted_forest_c,
+                orchard_name=orchard_name,
+                orchard_c=orchard_c,
+                logger=self.logger,
             )
+
             # Add to grid
-            self.set_grid(usle_c)
+            if usle_c is not None:
```
shartgring, line 508: usle_c would only be None when planted_forest is empty. So the calling of the method and writing the grid could be placed inside `if planted_forest: ...`

hydromt_wflow/workflows/landuse.py
shartgring: As mentoined in the other part of the code, an alternative is to use something like `if planted_forest: add_planted_forest_to_landuse(...)` when calling the workflow and making it required

tests/test_model_methods.py
```diff
@@ -846,6 +846,33 @@ def test_setup_lulc_sed(example_sediment_model, planted_forest_testdata):
...
+    example_wflow_model,
+    example_sediment_model,
+    globcover_gdf,
+    planted_forest_testdata,
+):
+    # Test for wflow sbm
+    example_wflow_model.setup_lulcmaps_from_vector(
+        lulc_fn=globcover_gdf,
+        lulc_mapping_fn="globcover_mapping_default",
+        lulc_res=0.0025,
```
shartgring, line 859: Also good to include a test where lulc_res=None just to be sure

shartgring (changes requested):
Hi Hélène! Nice feature, I will test it myself this week to make some nice LULC maps, which I luckily still have to do for a project. These are my comments based on reviewing only the code and not running it yet

hydromt_wflow/wflow.py
hboisgon: The WflowModel.maps comes from HydroMT core (dictionary of RasterDataArray or RasterDataset where resolution can be different). So officially it does not really exist for wflow ie we do not have a read_maps or write_maps method and I just write the files directly during the setup function).
I'm not sure if we want an official maps object though for Wflow. Staticgeoms is super useful for plotting or some processing within hydromt but not sure about the maps. And tables are official Wflow files ie wflow model does read them when running (eg rating curves of resrvoirs).

hydromt_wflow/workflows/landuse.py
hboisgon: So yes and no... When I do self.data_catalog.get_geodataframe(), the model bounds are already used to clip the gdf. We do intersection of the gdf so the bounds may still be bigger than the model.
The reason why I first prefer to use gdf.total_bounds is that I had some landuse shapefile that clearly came from a raster at very high resolution. So using a wrong bounding box would mean loosing a lot of info because then the grid is not aligned with the landuse shapefile.
Else I can also make this an argument of the function? Use the gdf or the model bbox. What do you think?

hydromt_wflow/workflows/landuse.py
hboisgon: Yes I hesitated and wanted to avoid too much code duplicate in wflow_sediment.py... but I agree it would still be the best place so I will shift it :)

hydromt_wflow/workflows/landuse.py
hboisgon: Good point because I'm already not fully sure when they came from or if I can find a ref again ^^'
This is actually true for many of the default values in hydromt, especially the parameters in the landuse mapping tables...

tests/test_model_methods.py
hboisgon: I did below when testing the wflow_sediment model :)

hydromt_wflow/wflow.py
shartgring: Good point! Let's just keep it like this then

tests/test_model_methods.py
shartgring: Oops missed that one

hydromt_wflow/workflows/landuse.py
hboisgon: I like it :)

hydromt_wflow/workflows/landuse.py
hboisgon: Found it :)

hboisgon:
Thanks for the useful review and comments @shartgring ! I think I adressed it all :) could you review again?

shartgring (approved):
Looks good to me! Only 1 thing is missing: I noticed that the LULC map (wflow_landuse) is not clipped to the basin, but to the bounding box. The parameter maps themselves are clipped correctly to the basin. So that is a small thing that is nice to address

Still I'll approve it and nice that it can be included in the release!

PR #323, opened 2025-01-22
python 39 support

### LINKED ISSUE #322: Drop support for python 3.9
### Kind of request

Removing existing functionality

### Enhancement Description

We now have the issue that python 3.9 is not tested or supported because of
pipe typing nodata_value: float | int = -9999 in [demand.py#L411](https://github.com/Deltares/hydromt_wflow/blob/main/hydromt_wflow/workflows/demand.py#L441) which is only allowed for python >=3.10.

3.9 support is being dropped for different packages but for hydromt-wflow I think it would be better to drop it once we migrate to hydromt v1 #275 and possibly wflow v1 #319 
So the above issue will be fixed for now.

### Use case

_No response_

### Additional Context

_No response_

hboisgon:
## Issue addressed
Fixes #322 
See also: https://github.com/conda-forge/hydromt_wflow-feedstock/pull/21

## Explanation
Fix and small release to fix support for python 3.9. Will be dropped again when moving to hydromt v1 / wflow v1
> in the end, decided to drop support already to avoid additional testing for python 39

## Checklist

### DIFF
--- a/.github/workflows/test_stable.yml
+++ b/.github/workflows/test_stable.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest" ] #, "macos-latest", "windows-latest"]
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.10', '3.11']
         include:
         - os: ubuntu-latest
           label: linux-64
--- a/.github/workflows/test_stable.yml
+++ b/.github/workflows/test_stable.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest" ] #, "macos-latest", "windows-latest"]
-        python-version: ['3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11']
         include:
         - os: ubuntu-latest
           label: linux-64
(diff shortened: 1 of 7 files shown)


hboisgon:
@savente93 can you help me with the failing py3.9 test? It fails as no cache is found and I have no idea how to fix it.
If this is too much work then we can prematurely drop python 3.9 support but I would like to do a minor release (0.7.1) and I'm not sure what would a number be if python 3.9 is officially dropped?

savente93:
I'll have a look when I get to work and see if I can fix it. If it's just about a cache in the workflow it should be fairly easy. is this urgent?

hboisgon:
After some discusssions with @savente93, decided to just drop support for python 39 already as it probably was not working for a while anyhow so no need for a major release.

hboisgon:
I'm still waiting to see if PR #320 can be part of the release. Once I know, I'll create the release tag. Could you already review @savente93 to see if I added all that is needed to upgrade to python 310? Thanks :)

savente93 (approved):
I think it looks good, although I will say that if you're dopping support for python 3.9 then you can actually keep the `X | Y` instead of `Union[X,Y]` and `X | None` instead of `Optional[X]` since those syntaxes were added in python 3.10 and I think it looks nicer. At least it doesn't require you to import the `Union` and `Optional` types. Though of course it is still valid, so if you preferer it that way you can keep the PR as is. I'll leave that up to you :)

PR #327, opened 2025-02-11
Fix formatting and typos

savente93:
## Explanation
This is just fixing a few typos and format mistakes in the code, no actual code has been changed. I'm doing this in prep to start work on the v1 adaptation as a gentle start. 

## Checklist


hydromt_wflow/workflows/waterbodies.py
```diff
@@ -632,7 +632,7 @@ def lakeattrs(
                 else:
                     logger.warning(
                         f"Rating data not available for lake {id}. \
-Using default Modified Puls Approach"
+Using default Modified Pulse Approach"
```
hboisgon, line 635:
```suggestion
Using default Modified Puls Approach"
```
It was good before :) (scientific term for the method name)

hboisgon (commented):
Thanks @savente93 for the updates! Nice way to start diving into hydromt_wflow :)
Seems though that the linting test fails with your modifications, could you update it?

Tests and docs are failing on cache restore, not sure why?  Could you check as well?
Anyway for fully passing tests we will need #330 to be merged but the docs should not be failing at this stage.

PR #334, opened 2025-02-26
Improve demand

hboisgon:
## Issue addressed
No issue created.

Adds two new methods for water demands:

- **setup_domestic_demand_from_population**
- **setup_irrigation_from_vector**

It also adds an extra argument to **setup_allocation_areas** to filter subbasins that would be too small (eg if a river meanders very closely to another admin area then very small allocation areas can be created).

## Explanation
I already encountered twice that local irrigated areas are in vector format rather than raster and we have a good method in core to treat vector data (grid_from_geodataframe with method fraction) so that avoids to loose too much data for a user that would first need to rasterize the map themselves and risk loosing info.

Secondly I think getting demand based on consumption per capita was already on the wishlist and it already poped up for the same study so I went ahead and added. In another version, I guess we may want to have these statistics stored in a admin area shapefile rather than a constant value but this will be a new feature for later.

## Checklist

I should have done it in another PR but while working on the same model, I realised that computing river depths in **setup_rivers** using *manning* or *gvf* methods was apparently broken for two years....
And while this PR is still open for review, I also added that parameters can be derived based on soil texture mapping with default values for InfiltCapSoil. I'll put the references (USDA, FAO) when we update the v1 docs together with the review of parameters for landuse.


hydromt_wflow/workflows/demand.py
```diff
@@ -602,3 +691,78 @@ def irrigation(
...
+    lai_threshold: float = 0.2,
+    logger=logger,
+):
+    """
+    Prepare irrigation maps for paddy and non paddy from geodataframe.
+
+    Parameters
+    ----------
+    da_irrigation: gpd.GeoDataFrame
+        Irrigation map
```
JoostBuitink, line 711:
```suggestion
    gdf_irrigation: gpd.GeoDataFrame
        Irrigation map
```

Instead of irrigation map, it should be irrigation shapefile or something, right?

hydromt_wflow/workflows/demand.py
```diff
@@ -602,3 +691,78 @@ def irrigation(
...
+    ds_like: xr.Dataset
+        Dataset at wflow model domain and resolution.
+
+        * Required variables: ['wflow_landuse', 'LAI']
+    irrigation_value: List[int]
+        Values that indicate irrigation in da_irrigation.
```
JoostBuitink, line 717:
```suggestion
        Values that indicate irrigation in gdf_irrigation.
```

hydromt_wflow/workflows/soilgrids.py
```diff
@@ -469,6 +471,10 @@ def soilgrids(
 producing quality-assessed soil information for the globe. SOIL Discussions, pp.1-37.
     https://doi.org/10.5194/soil-2020-65.
 
+    A ``soil_mapping`` table can optionnally be provided to derive parameters based
```
JoostBuitink, line 474:
```suggestion
    A ``soil_mapping`` table can optionally be provided to derive parameters based
```

tests/test_model_methods.py
```diff
@@ -662,6 +679,60 @@ def test_setup_rivers(elevtn_map, floodplain1d_testdata, example_wflow_model):
...
+        rivdph_method="gvf",
+        min_rivdph=1,
+        min_rivwth=30,
+        slope_len=2000,
+        smooth_len=5000,
+        river_routing="local-inertial",
+        elevtn_map="dem_subgrid",
+    )
+
+    assert "RiverDepth" in mod.grid
```
JoostBuitink, line 733: Lets say the manning method did produce a RiverDepth map, but the gvf method did not. Will this test capture that? I think it will not flag it right, since the RiverDepth is already part of the grid due to the previous method.

In addition, should we also do a similar test for the powlaw method?

JoostBuitink (changes requested):
Nice work and great additions! Looks good to me, just a couple of small typo's and questions, after that we are good to merge I think!

tests/test_model_methods.py
hboisgon: It's not the best test but if there is any error (eg missing dem_subgrid map which was the curretn issue) then the method itself will error. So the assert is not really the goal but justchecking that the method runs without errors. For powlaw we are testing when we build models as this is the default method. Is this test okay then? Or do you want me to test further with some values check or other?

hydromt_wflow/workflows/demand.py
hboisgon: Indeed! Forgot to replace when I copy/pasted :) I'll change it.

savente93:
Discussed this with Hélène, the one remaining comment is not to bad, I just added an explanation as a comment for future reference. Since Joost seems to be out of office, so I'll be merging this.

PR #335, opened 2025-02-28
release hydromt git dependency

savente93:
This was accidentally merged because I was a little trigger happy with merging. This reverts the bit that shouldn't have made it into main.


hboisgon (approved):
Thanks for fixing this @savente93 ! Looks good now :)

PR #336, opened 2025-03-03
Add support for inverse clipping

### LINKED ISSUE #112: inverse mask for model clipping
Add support for an inverse mask when clipping a model. This to essentially remove a subbasin upstream of a certain point from the main model, to allow for linking of two different model configurations. Option to do inverse masking should be parsed as a flag in the function arguments.

https://github.com/Deltares/hydromt_wflow/blob/4f2639259f7ffc2e4489a991cd0cefc02a1f6227/hydromt_wflow/wflow.py#L2153

JoostBuitink:
## Issue addressed
Fixes #112 

## Explanation
With a flag, a user can specify if the clipping needs te be done "normally" (only saving all upstream area given a location), or "inversely" (removing the upstream area given a location). This is useful to link for example several separate wflow models, that together cover a large river basin.

## Checklist


hboisgon (changes requested):
Thanks for creating and updating the PR @JoostBuitink.
I like your solution too :)

Before merging, it would still be good to update the docstrings and also implement this to the clip_forcing method. Can you still do this? Else I'm happy to take over too, just let me know!

JoostBuitink:
> Thanks for creating and updating the PR @JoostBuitink. I like your solution too :)
> 
> Before merging, it would still be good to update the docstrings and also implement this to the clip_forcing method. Can you still do this? Else I'm happy to take over too, just let me know!

Ah forgot about the docstrings, will do that! Related to the clip_forcing, I understood that it uses the extent of the grid (staticmaps) to determine how the forcing should be clipped. So I don't think there are changes required there, is that right? I think the same holds for the clip_states function, but let me know if changes are needed in these functions as well!

hboisgon (approved):
Of course you are right for forcing! Sorry I forgot how it was working... So all good now!
I was hesitating to add this change in the version 1 release. What do you think? If version 1 then we need to wait before we can merge.

JoostBuitink:
> Of course you are right for forcing! Sorry I forgot how it was working... So all good now! I was hesitating to add this change in the version 1 release. What do you think? If version 1 then we need to wait before we can merge.

No problem! I think this can be merged now (since it is also a very minor code change), such that for v1 changes it will also be included. Or do I overlook something?

PR #339, opened 2025-03-11
feat: enable use of pixi as package manager

### LINKED ISSUE #338: Support pixi as package manager
### Kind of request

Adding new functionality

### Enhancement Description

add a pixi.toml or a pixi section in `pyproject.toml` with the dependencies and tasks so that we can install a development environment using pixi as well 

### Use case

Conda is slow and has reproducibility issues. Enabeling the use of pixi could solve this for those that want to use it. It also enables running of "scripts" like building the docs cross platform which is 

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #338 

## Explanation
I've tried to organize the environments as sensibly and as close to the existing groups as possible. I've also added some tasks that weren't documented (as explicitly) before which should be useful. 

## Checklist
- [x] Branch is up to date with `v1`

Tests will not pass this is expected, since the point is to come up to spec with v1. so for now this will be inevitable


pixi.toml
deltamarnix, line 1: I would remove pixi.toml if everything is in pyproject.toml

pyproject.toml
```diff
@@ -114,3 +114,23 @@ filterwarnings = [
...
+hydromt_wflow = { path = ".", editable = true }
+
+[tool.pixi.environments]
+default = { solve-group = "default" }
+dev = { features = ["dev"], solve-group = "default" }
+doc = { features = ["doc", "examples", "extra"], solve-group = "default" }
+examples = { features = ["examples"], solve-group = "default" }
+extra = { features = ["extra"], solve-group = "default" }
+full = { features = ["full", "io", "extra", "dev", "test", "doc", "examples"], solve-group = "default" }
+io = { features = ["io"], solve-group = "default" }
```
deltamarnix, line 132: I think you don't need so many environments.
I miss the python versions and the solve groups should be set to the different python versions. See my updated pyproject.toml file in this PR: https://github.com/Deltares/hydromt/pull/1125

.gitignore
```diff
@@ -136,3 +136,7 @@ environment.yml
 
 # dask
 dask-worker-space/
+
+# pixi environments
+.pixi
```
deltamarnix, line 141: I see that the `.vscode` folder is ignored. I would recommend not doing that.

pyproject.toml
```diff
@@ -11,18 +11,18 @@ authors = [
...
-    "pandas",
-    "pyflwdir>=0.5.7",
-    "pyproj",
-    "scipy",
-    "shapely",
-    "toml",
-    "xarray",
+  "dask",
+  "geopandas>=0.10",
+  "hydromt>=v1.0.0",
```
deltamarnix, line 16:
```suggestion
  "hydromt>=v1.0.0,<2",
```

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
   "Intended Audience :: Developers",
   "Intended Audience :: Science/Research",
   "Topic :: Scientific/Engineering :: Hydrology",
-  "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
+  "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3)",
```
deltamarnix, line 34: What?! This package is under GPL license?

pyproject.toml
```diff
@@ -114,3 +114,124 @@ filterwarnings = [
...
+
+[tool.pixi.feature.py312.dependencies]
+python = "3.12.*"
+
+[tool.pixi.feature.extra.dependencies]
+gdal     = "*"
+pcraster = "*"
+
+[tool.pixi.environments]
+default = { solve-group = "default" }
```
deltamarnix, line 139: I would recommend, just like hydromt core, to make the `default` environment the one that is for development. Remove `dev`. I think you can do with a whole lot less.

Maybe just:
- `default` (solve group py310)
- `py310`
- `py311`
- `py312`
- `py313`

There seems to be no use for minimal and slim in this case.
Why all these different development options? I don't think anyone is ever going to develop with different python versions.

pyproject.toml
```diff
@@ -114,3 +114,124 @@ filterwarnings = [
...
+clean = { depends-on = [
+  "clean-dist",
+  "clean-docs-generated",
+  "clean-docs-build",
+  "clean-docs-exmaples",
+] }
+clean-dist = { cmd = ["rm", "-rf", "dist"] }
+clean-docs-generated = { cmd = ["rm", "-rf", "docs/_generated"] }
+clean-docs-build = { cmd = ["rm", "-rf", "docs/_build"] }
+clean-docs-exmaples = { cmd = ["rm", "-rf", "docs/examples"] }
```
deltamarnix, line 224:
```suggestion
  "clean-docs-examples",
] }
clean-dist = { cmd = ["rm", "-rf", "dist"] }
clean-docs-generated = { cmd = ["rm", "-rf", "docs/_generated"] }
clean-docs-build = { cmd = ["rm", "-rf", "docs/_build"] }
clean-docs-examples = { cmd = ["rm", "-rf", "docs/examples"] }
```

pyproject.toml
```diff
@@ -114,3 +114,124 @@ filterwarnings = [
...
+  "extra",
+  "dev",
+  "test",
+  "doc",
+  "examples",
+  "py312",
+], solve-group = "py312" }
+[tool.pixi.tasks]
+
+install-pre-commit = "pre-commit install"
```
deltamarnix, line 182: Maybe also add an `install = { depends-on = ["install-pre-commit"] }`

deltamarnix (changes requested):
You might want to remove some more environments that are not necessary

pyproject.toml
savente93: apparently 🤷‍♀️

pyproject.toml
savente93: good point, I'll do that

pyproject.toml
```diff
@@ -114,3 +118,122 @@ filterwarnings = [
...
+
+[tool.pixi.feature.py310.dependencies]
+python = "3.10.*"
+
+[tool.pixi.feature.py311.dependencies]
+python = "3.11.*"
+
+[tool.pixi.feature.py312.dependencies]
+python = "3.12.*"
+
```
deltamarnix, line 137: Let's add 3.13?

pyproject.toml
```diff
@@ -114,3 +118,122 @@ filterwarnings = [
...
+[tool.pixi.feature.extra.dependencies]
+gdal     = "<=3.8" # pin is necesary to avoid dependency conflict with numba re numpy 2.2.4
+pcraster = "*"
```
deltamarnix, line 140: Moving this to pixi dependencies makes that a user consuming hydromt_wflow no longer gets gdal and pcraster. Would all extras still work for them?

pyproject.toml
```diff
@@ -114,3 +118,122 @@ filterwarnings = [
...
+
+# clean
+clean = { depends-on = [
+  "clean-dist",
+  "clean-docs-generated",
+  "clean-docs-build",
+  "clean-docs-examples",
+] }
+
+clean-dis            = { cmd = ["rm", "-rf", "dist"] }
```
deltamarnix, line 222:
```suggestion
clean-dist            = { cmd = ["rm", "-rf", "dist"] }
```

pyproject.toml
savente93: good one, my bad

pyproject.toml
savente93: so, upon quite some investigation the `extra` group contained a mix of `pypi` and `conda` dependencies, which the `make_env.py` would separate out with the use of the `[tool.make_env]` section so this on it's own would never have worked (I suspect nobody ever tried to install it as `pip install .[extra]`. So in a way you're right, but it was also never correct to begin with. I'll see if I can convince make_env to put them back

savente93: should be good now

deltamarnix (approved):
🥳

PR #349, opened 2025-03-26
Masking grid with subcatch on set instead of on write

### LINKED ISSUE #325: Masking with subcatch in grid

### Reproducible Example

Run the test_model_methods.py::test_setup_lulc_vector and examine the example_wflow_model after calling the setup method.

### Current behaviour

Not a real bug but currently masking of variables in grid with the basin boundaries happens in ``write_grid`` rather then in the ``setup_`` methods. I think it would be better if it hapened during the ``set_grid`` method so that we have a more conform model in python and not only after/when writting.

### Desired behaviour

See above.

### Additional context

_No response_

savente93:
## Issue addressed
Fixes #325 

## Explanation
Previously the grid was masked to the subcatch but only on write, meaning that if you read and wanted to do comparisons or make statistics you'd either have to do the masking manually, or you got wrong results. This is quite counterintuitive, so we moved the masking from the writing to the setting of the grid. 

Some extra contingencies had to be added since at the time of setting any grid, the subcatch might not exist (yet) or be not relevant, so in those cases we just do nothing. 

The clipping is applied again on the write to preserve previous behaviour. perhaps this is unnecessary, but I did it to be sure. This can easily be removed in necessary. 

I also fixed a new mistakes in the error reporting code I noticed while I was busy, but these are non-functional so it's not a big deal. 

## Checklist

### DIFF
--- a/hydromt_wflow/utils.py
+++ b/hydromt_wflow/utils.py
@@ -150,6 +150,7 @@ of the config.
                     # Filter and reduce full_index based on mask
                     full_index = full_index.where(mask != mask.raster.nodata, 0)
                     full_index.attrs.update(_FillValue=0)
+
                     mask_index = full_index.values.flatten()
                     mask_index = mask_index[mask_index != 0]
                     # idx corresponding to the wflow index
@@ -316,3 +317,41 @@ def get_grid_from_config(
             da = grid[var_name] * scale + offset
 
     return da
+
+
+def mask_raster_from_layer(ds_grid: xr.Dataset, layer_name: str) -> xr.Dataset:
+    """Mask the data in the supplied grid based if the value in one of the layers.
+
+        This for example can be used to mask a grid based on subcatchment data.
+        All data in the rest of the data variables in the dataset will be set to
+        the `raster.nodata` value except for boolean variables which will be set to
+        False. If the supplied grid is not an xarray dataset, or does not contain
+        the correct layer, it will be returned unaltered.
+
+        The layer supplied can be either boolean or numeric. Array elements where the
+        layer is larger than 0 (after type conversion) will be masked.
+
+    Parameters
+    ----------
+        ds_grid (xr.Dataset):
+            The grid containing both the data that should be masked and a layer that
+            will be used to determine the mask
+        layer_name (string):
+            name of the layer that the grid will be masked to. Values can be boolean or
+            numeric. Places where this layer is greater than 0 will be masked in the
+            other data variables
+
+    Returns
+    -------
+        xr.Dataset: The grid with all of the data variables masked.
+    """
+    if isinstance(ds_grid, xr.Dataset) and layer_name in ds_grid.data_vars:
+        mask = ds_grid[layer_name] > 0
+        for var in ds_grid.data_vars:
+            # nodata is required for all but boolean fields
+            if ds_grid[var].dtype != "bool":
+                ds_grid[var] = ds_grid[var].where(mask, ds_grid[var].raster.nodata)
+            else:
+                ds_grid[var] = ds_grid[var].where(mask, False)
+
+    return ds_grid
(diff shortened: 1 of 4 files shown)


hydromt_wflow/wflow.py
```diff
@@ -4567,13 +4567,21 @@ def write_grid(
...
-        mask = ds_out[self._MAPS["basins"]] > 0
-        for v in ds_out.data_vars:
-            # nodata is required for all but boolean fields
-            if ds_out[v].dtype != "bool":
-                ds_out[v] = ds_out[v].where(mask, ds_out[v].raster.nodata)
+        ds_out = self._mask_grid_to_basins(ds_out)
+
         ds_out.to_netcdf(fn, encoding=encoding)
 
+    def _mask_grid_to_basins(self, ds_out):
```
hboisgon, line 4574: Nice! maybe you can `basin_layer_name` as an argument and move this function to utils rather than model?

hydromt_wflow/wflow.py
```diff
@@ -4567,13 +4567,21 @@ def write_grid(
         if not isdir(dirname(fn)):
             os.makedirs(dirname(fn))
         self.logger.info(f"Write grid to {fn}")
-        mask = ds_out[self._MAPS["basins"]] > 0
-        for v in ds_out.data_vars:
-            # nodata is required for all but boolean fields
-            if ds_out[v].dtype != "bool":
-                ds_out[v] = ds_out[v].where(mask, ds_out[v].raster.nodata)
+        ds_out = self._mask_grid_to_basins(ds_out)
```
hboisgon, line 4570: The delineation of the basin is actually the very first thing that should be done based on the region. unless we change this behavior in wflow v1, masking on write should then not be needed anymore :)

tests/test_model_class.py
```diff
@@ -54,9 +54,11 @@ def _compare_wflow_models(mod0, mod1):
                     else f"{map1.dtype} instead of {map0.dtype}"
                 )
                 # Check on nodata
+                # hilariously np.nan == np.nan returns False, hence the additional check
```
hboisgon, line 57: funny indeed ^^ we barely use nan as a nodata value so guess that's why so far we did not see this

hboisgon (commented):
Looks good Sam thanks for this!
I have a few suggestions but I'm also happy to leave as you have done now if you prefer.

hydromt_wflow/wflow.py
```diff
@@ -28,6 +28,8 @@
 from pyflwdir import core_conversion, core_d8, core_ldd
 from shapely.geometry import box
 
+from hydromt_wflow.utils import mask_raster_from_layer
+
```
hboisgon, line 32: If you look 1 and 3 lines below we import .utils or things from utils a lot here. Is it best to use relative import or hydromt_wflow ie .utils or hydromt_wflow.utils?
I'm not sure what is the best here but would be happy if you could harmonise with what you think is best :)

hydromt_wflow/utils.py
```diff
@@ -316,3 +317,41 @@ def get_grid_from_config(
             da = grid[var_name] * scale + offset
 
     return da
+
+
+def mask_raster_from_layer(ds_grid: xr.Dataset, layer_name: str) -> xr.Dataset:
+    """Mask the data in the supplied grid based if the value in one of the layers.
```
hboisgon, line 323:
```suggestion
    """Mask the data in the supplied grid based on the value in one of the layers.
```

hboisgon (approved):
Thanks @savente93! Just one last remark if maybe we could harmonise import from utils in wflow.py but this is small so I already approve so you can merge later.
Oh and I forgot but would be good to update in the changelog that masking with subcatchment will happen in set_grid and not write_grid anymore ;) might also be worth adding to the docstring of the ``set_grid`` function

PR #358, opened 2025-04-08
wradlib interpolation_classes dict only created when wradlib is installed

### LINKED ISSUE #354: Bug with not required wradlib library

### Reproducible Example

Check the latest conda forge install test where wradlib is not installed.

### Current behaviour

Error with import of wrb in the interpolation_classes dict

https://github.com/Deltares/hydromt_wflow/blob/82a66fe05642716d980eed3b64159ac8f30ea5ca/hydromt_wflow/workflows/forcing.py#L10-L31

### Desired behaviour

Define the dict only if HAS_WRADLIB

### Additional context

_No response_

Tjalling-dejong:
## Issue addressed
Fixes #354 

## Explanation
I added an if-statement to check if wradlib is installed and then create the interpolation_classes dict

## Checklist

### DIFF
--- a/hydromt_wflow/workflows/forcing.py
+++ b/hydromt_wflow/workflows/forcing.py
@@ -18,6 +18,7 @@ logger = logging.getLogger(__name__)
 
 __all__ = ["pet", "spatial_interpolation"]
 
+
 if HAS_WRADLIB:
     interpolation_classes = {
         "nearest": {"obj": wrl.ipol.Nearest, "args": None},
--- a/hydromt_wflow/workflows/forcing.py
+++ b/hydromt_wflow/workflows/forcing.py
@@ -18,17 +18,17 @@ logger = logging.getLogger(__name__)
 
 __all__ = ["pet", "spatial_interpolation"]
 
-
-interpolation_classes = {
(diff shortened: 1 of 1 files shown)


Huite:
(Random drive by comment:) Easiest solution here is to move the import statement and the dictionary definition in the function., since it's only used in this specification function.

In cases an optional module is needed in more places, I like to use this personally:

```python
class MissingOptionalModule:
    """Presents a clear error for optional modules."""

    def __init__(self, name):
        self.name = name

    def __getattr__(self, name):
        raise ImportError(f"{self.name} is required for this functionality")
```

Then in use:

```python
try:
    import wradlib
except ImportError
    wradlib = MissingOptionalModule("wradlib")
```

hboisgon:
Random drive by indeed ^^
I think the error message in the function is clear (also mentions in this case the name of the function) so I don't think we need a class yet.
It is true that wradlib is only used in one function so far so we could move the interpolation dictionnary into the function directly after the wradlib check has been done. @Tjalling-dejong can you update this?

hboisgon (approved):
LGTM!

PR #361, opened 2025-04-08
Pixi support

dalmijn:
## Issue addressed
Fixes pixi support

## Explanation
Added pixi to the pyproject.toml, adjusted the workflows accordingly

## Checklist

No.


.github/workflows/docs.yml
```diff
@@ -1,104 +1,71 @@
+---
 name: Build Documentation
 
 on:
   workflow_dispatch:
   push:
     branches: [main]
     paths:
-      - tests/*
-      - hydromt_wflow/*
```
savente93, line 9: I think we do want to generate the docs if the code gets adjusted.

.github/workflows/docs.yml
```diff
@@ -1,104 +1,71 @@
...
 jobs:
   # Build docs on Linux
-  Docs:
+  docs:
     defaults:
       run:
         shell: bash -l {0}
+    env:
+      branch_name: ${{ github.head_ref || github.ref_name }}
+    timeout-minutes: 30
```
savente93, line 29: I think 30 min is very high. 10-15 min should be enough

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
+name: Test Package
+
+on:
+  push:
+    branches: [main]
+    paths:
```
savente93, line 6: also add this file

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+      - .github/workflows/docs.yml
+
+jobs:
+  test:
+    defaults:
+      run:
+        shell: bash -l {0}
+    env:
+      branch_name: ${{ github.head_ref || github.ref_name }}
+    timeout-minutes: 30
```
savente93, line 26: Think the timeout can be way lower

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
```
savente93, line 31: it's missing windows

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          pixi run -e test-full-py${{ matrix.python-version }} test-cov
+
+      - name: Upload coverage to codecov.io
+        if: matrix.python-version == '311'
+        uses: codecov/codecov-action@v5
+        with:
+          files: ./coverage.xml
+          token: ${{ secrets.CODECOV_TOKEN }}
+          verbose: true
```
savente93, line 66: This is not necessary, sonarcube will do this which I'll add in my PR for that

.github/workflows/test_conda.yml
```diff
@@ -0,0 +1,44 @@
+name: Conda dependency check
```
savente93, line 1: it's not entirely clear to me what this workflow does. It's called conda, but uses pixi, and it says it checks the dependencies, but it just installs the deps from the lockfile, so I don't think this does anything.

pyproject.toml
```diff
@@ -1,20 +1,40 @@
+## Build system
 [build-system]
 requires      = ["flit_core >=3.4.0,<4"]
 build-backend = "flit_core.buildapi"
 
+[tool.flit.sdist]
+include = ["hydromt_wflow"]
+exclude = ["docs", "examples", "envs", "tests", "binder", ".github"]
```
savente93, line 8: we don't have "envs" or "binder" folders as far as I'm aware

pyproject.toml
```diff
@@ -24,70 +44,95 @@ dependencies = [
...
 ]
 examples = [
-  "jupyterlab",
   "cartopy",
+  "jupyterlab",
   "nbconvert==7.13.1", # Pin as of (02-2024); 7.14 onwards causes issues with the progressbar in example notebooks. See issue #233 for more info (https://github.com/Deltares/hydromt_wflow/issues/233).
   "notebook",
 ]
+test = [
+  "pytest>=2.7.3",
```
savente93, line 74: I think you're missing `pytest-timeout`

pyproject.toml
```diff
@@ -24,70 +44,95 @@ dependencies = [
...
+    # Don't complain about missing debug-only code:
+    "def __repr__",
+    "if self\\.debug",
+
+    # Don't complain if tests don't hit defensive assertion code:
+    "raise AssertionError",
+    "raise NotImplementedError",
+
+    # Don't complain if non-runnable code isn't run:
+    "if 0:",
```
savente93, line 115: this one I disagree with. I don't want non-runnable code in the repo

pyproject.toml
```diff
@@ -96,21 +141,94 @@ docstring-code-format      = true
...
+# Clean up (some of) the temporary files
+clean = { depends-on = [
+  "clean-env-files",
+  "clean-bin",
+  "clean-dist",
+  "clean-docs-build",
+  "clean-docs-example",
+  "clean-docs-gen",
+] }
+clean-env-files = { cmd = ["rm", "-f", "environment.yml"] }
```
savente93, line 190: we removed `make_env.py` so I don't think this is necessary anymore

pyproject.toml
```diff
@@ -96,21 +141,94 @@ docstring-code-format      = true
...
+clean-docs-gen = { cmd = ["rm", "-rf", "docs/_generated"] }
+clean-docs = { depends-on = [
+  "clean-docs-build",
+  "clean-docs-example",
+  "clean-docs-gen",
+] }
+
+## Dependencies
+[tool.pixi.dependencies]
+h5py = "*"
```
savente93, line 204: this one is new, what does it do?

pyproject.toml
```diff
@@ -96,21 +141,94 @@ docstring-code-format      = true
...
+## Define the environments
+[tool.pixi.environments]
+default = { features = ["py310", "dev", "docs", "examples", "extra", "test"] }
+docs = { features = ["docs", "examples", "extra"] }
+full = { features = ["extra"] }
+slim = { features = [] }
+test-slim-py311 = { features = ["py311", "test"] }
+test-slim-py310 = { features = ["py310", "test"] }
+test-full-py311 = { features = ["py311", "extra", "test"] }
+test-full-py310 = { features = ["py310", "extra", "test"] }
```
savente93, line 234: slim in this situation is not neceesay, since it is equivalent to `--no-default-features` so I think we can simply this

savente93 (changes requested):
Looks good in general but I do have a couple of questions

.github/workflows/docs.yml
dalmijn: But not the tests I would say

pyproject.toml
dalmijn: Didnt know that flag existed, fair enough

.github/workflows/test_conda.yml
dalmijn: To test whether the newest conda deps are compatible, however it forgot to add the manual update to the newest packages.

pyproject.toml
dalmijn: I still leave it in, as it might be more convenient

pyproject.toml
dalmijn: Well, it's not used at this point. Put it in anyway

pyproject.toml
dalmijn: Deleted it.

pyproject.toml
dalmijn: Fair enough.

.github/workflows/docs.yml
savente93: yeah that's true

savente93:
will be happy to rereview once the windows versions also run successfully

savente93 (approved):
nice work Brendan!

PR #363, opened 2025-04-08
fix hanging due to open netcdf files

### LINKED ISSUE #271: Github workflows for tests and doc often hang with python 3.11 on linux

### Reproducible Example

NA

### Current behaviour

Github tests and docs workflows often hang on linux with python 3.11. We should investigate why this behavior happens and improve our tests.

### Desired behaviour

No more hanging

### Additional context

_No response_

LuukBlom:
## Issue addressed
Fixes #271 

## Explanation
Replaced all occurences of `xr.open_dataset` and `xr.open_mfdataset` with context managers.
This ensures all .nc files are closed when no longer needed. 

Note that this is still a quick fix and did not resolve the hanging workflow issue.
hydromt.DataCatalog often returns a dataset with an open file handle and Im not sure if that causes problems here or if this fix solves it, especially because the bug from #271 is difficult to reproduce in general.

As a band-aid solution, I added [retries](https://github.com/nick-fields/retry) to the relevant gh actions workflows so you dont have to manually retrigger the workflow.

## Checklist


savente93 (changes requested):
All the tests are failing, and it seems that the docs actually still hung with this implementation, can you recheck it? (feel free to reach out if you want help with it)

LuukBlom:
Ah oops. Should be fixed now! Dont know what happened/how that fixture worked locally, but should be good to go now.

hboisgon (approved):
Thanks Luuk! Nice that the retry for the docs are automated!
I still wonder about the timeout for the notebooks? It's not nice but it may still be better to have an error else I'm not sure we will notice if the notebooks timed out during building the docs meaning that they would then be incomplete.

I'll let you decide on this and approve your changes for now.

LuukBlom:
> Thanks Luuk! Nice that the retry for the docs are automated! I still wonder about the timeout for the notebooks? It's not nice but it may still be better to have an error else I'm not sure we will notice if the notebooks timed out during building the docs meaning that they would then be incomplete.
> 
> I'll let you decide on this and approve your changes for now.

Okay, I'll merge this for now then!

I dont really see another good option to automatically notice / retry when a workflow is hanging. But agree its not a nice solution
If at some point there is a problem with the timeout, we can increase/remove it by increasing/removing the `nb_timeout` variable in `conf.py`

PR #372, opened 2025-04-09
change Uniont types to | notation

savente93:
## Explanation

The failing quality gate of sonarqube on main was really bothering me and it seemed to be entirely due to using `Union` instead of the new `|` syntax in python 3.10, so I fixed those. With this, the sonar workflow should also be green again on main. Sadly sonarqube doesn't show which issues it will close, but since there are no issues raised in the touched code, I think it should be fine. The coverage needs to be improved at some point but I don't want this PR to get too big so I'll leave that for now. 

## Checklist

Should probably be merged after release is done

### DIFF
--- a/hydromt_wflow/pcrm.py
+++ b/hydromt_wflow/pcrm.py
@@ -6,7 +6,6 @@ import os
 import tempfile
 from os.path import basename, dirname, isdir, isfile, join
 from pathlib import Path
-from typing import Union
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


dalmijn (approved):
Nice Sam! 
As far as I can see, your changes are all good.
We need to however take a careful look at the docstrings in the future.

dalmijn:
Agreed!

PR #377, opened 2025-04-14
release nbconvert pin

savente93:
## Explanation
Dependabot wanted to update the nbconvert version, which was pinned because of a specific issue. I checked locally and the issue seems to be gone, so I think we can remove the pin completely. 

## Checklist


dalmijn (approved):
LGTM.

PR #378, opened 2025-04-14
Add python to slim, full, and docs.

deltamarnix:
Add solve-groups to simplify pixi solving.

## Explanation
Making sure that python is installed in full, slim, and docs.
And add solve-groups to ensure that the different environments have the same versions.

## Checklist


savente93 (approved):
:D

savente93 (approved):
LGTM

savente93 (commented):
I just noticed that the `full` flavors don't have the dev, docs and examples included. Should they?

deltamarnix:
> I just noticed that the `full` flavors don't have the dev, docs and examples included. Should they?

I don't even think we need `full`. pixi environments are only there for running specific pixi tasks. So maybe we can get rid of them if they are not used in a docker environment or github action.

savente93:
I'd be okay with that

savente93:
I'll be merging this now to make sure we are correct when running the tests. We can discuss looking at whether we will need the `full` environments separately when Marnix returns from leave.

PR #387, opened 2025-04-17
Preserve existing structure of toml config files on write

### LINKED ISSUE #232: Improve behavior of `setup_constant_pars`
### Kind of request

### Enhancement Description

Currently the setup_constant_pars is adding maps with uniform values to the staticmaps file. Currently, there is also no entry added to the wflow TOML configuration file, but implicitly taken from the "reference" TOML file. This means that, if someone would like to add an additional constant parameter, the TOML file won't be updated correctly. 

Furthermore, to improve clarity and ease-of-use, it would be good if - in stead of a layer in the staticmaps - the value gets added to the wflow TOML configuration file, see the example below:

```toml
[input.vertical]
ksathorfrac.value = 100
```

### Use case

_No response_

### Additional Context

Currently, If passing the required information to the config object (through `parameter.value`), the TOML gets written in the following manner:

```toml
[input.vertical.ksathorfrac]
value = 100
```

While this is file for a limited number of constant parameters, I would prefer if it gets added like in the "Enhancement description" above. 

I did some work related to this in this branch: https://github.com/Deltares/hydromt_wflow/tree/toml_naming

Tjalling-dejong:
## Issue addressed
Fixes #232 

## Explanation
It turns out that `tomlkit` is quite a buggy implementation so this ended up taking quite long but we got there in the end. The rule we ended up with in the end is that when a value is set, it will be added to the most specific existing table in the config, if this does not exist it will be added to the root. This does mean that if for example the config file should have an input block like so:

```toml
[input.static]
foo = "bar"
mew = "bark"
```

then the table should be added to the toml document before hand even if there are no keys for it yet like so:

```toml
[input.static]
```

Then subsequent sets will be added to this table 
## Checklist

### DIFF
--- a/hydromt_wflow/utils.py
+++ b/hydromt_wflow/utils.py
@@ -231,7 +231,7 @@ def _get_key_values_to_move(
 
 
 def _update_config_dict_with_value_keys(
-    keys_values_to_move: List[tuple[str, str]], data: dict[str, str]
+    keys_values_to_move: List[tuple[str, Any]], data: dict[str, str]
 ):
     for parent, value in keys_values_to_move:
         parents, parent = parent.rsplit(".", maxsplit=1)
--- a/hydromt_wflow/utils.py
+++ b/hydromt_wflow/utils.py
@@ -1,12 +1,12 @@
 """Some utilities from the Wflow plugin."""
 
 import logging
-from functools import reduce
 from os.path import abspath, dirname, join
 from pathlib import Path
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 import numpy as np
+import tomlkit
 import xarray as xr
 from hydromt.io import open_timeseries_from_table
 from hydromt.vector import GeoDataArray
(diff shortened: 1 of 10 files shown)


hydromt_wflow/wflow.py
```diff
@@ -5692,12 +5693,133 @@ def set_tables(self, df, name):
...
+
+        args = list(args)
+        value = args.pop(-1)
+        keys = reduce(lambda acc, arg: [*acc, *arg.split(".")], args, [])
+
+        # if we try to set dictionaries as values directly tomlkit will messup the
+        # key bookkeeping, resulting in invalid toml, so instead
+        # if we see a mapping, we go over it recursively
+        # and manually add all of it's keys, which does work for some reason
+        if isinstance(value, (dict, tomlkit.items.Table)):
```
deltamarnix, line 5788: Would it also work if you would test for AbstractTable? I see that InnerTable is not covered otherwise.

hydromt_wflow/wflow.py
```diff
@@ -5692,12 +5693,133 @@ def set_tables(self, df, name):
...
+        for idx in range(len(keys)):
+            if idx != len(keys) - 1:
+                remaining_key = tomlkit.key(keys[idx:])
+            else:
+                remaining_key = keys[idx]
+
+            if keys[idx] not in current or not isinstance(current[keys[idx]], dict):
+                break
+
+            current = current[keys[idx]]
```
deltamarnix, line 5814: Is all this code still needed? I have to check, but I'm guessing that the recursive functionality on 5790 is already resolving all of your issues. `set_config` will actually only ever receive a single key to resolve. It looks like the function is a bit too superfluous for what it should do.

deltamarnix (changes requested):
Pleasea take a look at `set_config`, perhaps there is some redundant code. I will also check from my side.

hydromt_wflow/wflow.py
```diff
@@ -4876,8 +4877,9 @@ def upgrade_to_v1_wflow(self):
         The other components stay the same.
         This function should be followed by write_config() to write the upgraded file.
         """
+        self.read()
```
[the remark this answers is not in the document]
savente93: it is required, otherwise the test fails, which sort of makes sense, since if the model is not loaded and you modify in your config where geoms are located/what they are called it makes sense you cannot find them afterwards.

hydromt_wflow/utils.py
```diff
@@ -200,42 +201,48 @@ def read_csv_results(
...
 
     Returns
     -------
-    value : any type
+    value : Any
         dictionary value
+
+    Examples
+    --------
+    >> # self.config = {'a': 1, 'b': {'c': {'d': 2}}}
```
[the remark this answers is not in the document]
savente93: yeah true, this was in the old example and I didn't update it. The only thing that `set_config` usually initializes the default config which is a little overwhelming for the example, I'll see if I can craft something more consistent.

hydromt_wflow/wflow.py
savente93: Good point

hydromt_wflow/wflow.py
savente93: The code on 5790 is indeed to make sure we only try to set literals as values since otherwise things go awry with tomlkit. this code is to make sure we choose the most specific header for the key we are trying to set, so they do different things. I don't see how this can be reduced since they operate on different data

hydromt_wflow/wflow.py
```diff
@@ -5692,12 +5693,133 @@ def set_tables(self, df, name):
...
+        """
+        self._initialize_config()
+        if len(args) < 2:
+            raise TypeError("set_config() requires a least one key and one value.")
+        if not all([isinstance(part, str) for part in args[:-1]]):
+            raise TypeError("All but last argument for set_config must be str")
+
+        args = list(args)
+        value = args.pop(-1)
+        keys = reduce(lambda acc, arg: [*acc, *arg.split(".")], args, [])
```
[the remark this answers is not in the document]
savente93: I find nested list comprehensions harder to read but you can argue that based on taste. Although I'm a little unclear now on whether we think using reduce is good or not, since you initially introduced the `setdefault` reduce one liner. Do we like the use or reduce or not?

deltamarnix (approved):
We've checked the edge cases. It's not allowed to have quoted dots in a key. And wflow doesn't have them, so we're good.

Issue #351, opened 2025-03-28
Update Documentation


alimeshgi:
### HydroMT-Wflow version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt_wflow)

### Kind of issue

Docs are missing

### Documentation problem

We need to review and update the documentation to reflect the new changes introduced in version 1.

deltamarnix:
Don't forget to also update the yaml config files in the new format. Make it easy for the user to find the template.

deltamarnix:
For who ever is picking this up, I'm currently looking at the navbar.

Issue #332, opened 2025-02-25
Main issue: hydromt_wflow release 1.0


hboisgon:
### Enhancement Description

This issue is the main issue to recap the steps and wishes for a hydromt_wflow version 1 release.
In this release, we would like to take the steps to support:

- Wflow version 1
- HydroMT version 1

### Required developments

1. Create a small release to support hydromt v0.10.1

- [x] Release xarray and numpy pin #330 
- [x] Code formatting #327 
- [x] Finalise #334 
- [x] Finalise a first version of the precipitation interpolation #312 

2. HydroMT version 1 support

- [x] move to hydromt version 1 model object
- [ ] check if some of the functions can be updated to use core methods (especially for grid) #2
- [x] allow to clip wflow models #317 + inverse clipping
- [x] review buffer values (number of cells vs distance in meter)

3. Wflow version support

- [x] update to the new TOML structure and wflow parameters names #319 
- [x] support updating from pre-v1 TOML structure to v1 TOML structure #319 
- [x] better support updating of parameters names in the TOML file (user can defined names in the staticmaps nc and be linked to TOML) #232 
- [x] add or update parameters that have been added or changed in Wflow #309 #331 

4. Documentation / testing

- [ ] Update documentation and tests with the new changes #351 
- [x] Improve installation guide #324 

### Nice to have
Issues that are not required for a working support of hydromt and wflow versions 1 but good to have for hydromt_wflow itself (more or less in order of importance).

- [x] Masking with subbasins in self.grid #325 
- [ ] Clarify dependencies between setup methods #125 
- [x] Fix hanging tests and docs #271 
- [ ] Preserve TOML file structure #79 
- [ ] Review default values in the different setup methods #106 
- [ ] Review default values in the landuse mapping tables #84 
- [ ] Restructure docs #250

JoostBuitink:
Maybe good to add to the nice-to-have-list (since I see clipping is mentioned in the hydromt-v1 related issues), is support for inverse clipping (e.g. removing an upstream area from a wflow model). This is only a very minor change (see this branch https://github.com/Deltares/hydromt_wflow/compare/main...inverse_clipping), but would be nice to have in the main branch as well at some stage!

hboisgon:
[...]

PR #389, opened 2025-04-18
fix: change pixi global installation

### LINKED ISSUE #324: Improve installation guide in the docs

- [x] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt_wflow)

### Kind of issue

Docs are unclear

### Location of the documentation

https://deltares.github.io/hydromt_wflow/latest/getting_started/installation.html

### Documentation problem

Installation through mamba or conda may give some warnings like:

`warning  libmamba 'repo.anaconda.com', a commercial channel hosted by Anaconda.com, is used.`

It might be worth to emphasize the use of only conda-forge as a channel due to licensing, and encourage users to use miniforge. Also, the link to the HydroMT Core docs is broken

Additionally, HydroMT Core encourages the use of pixi, while HydroMT Wflow uses mamba/conda. Should this be addressed/mentioned before it might raise confusion?

### Suggested fix for documentation

- Adding a tip or warning to use miniforge which only has the conda-forge channel installed?
- Mention something about Pixi?

shartgring:
## Issue addressed
Fixes #324

## Explanation
For global installation the user must install and expose `hydromt` instead of `hydromt_wflow`

## Checklist

@savente93 specifying `pixi global install hydromt --with hydromt_wflow` results in hydromt version 1.1.0 with hydromt_wflow version 0.2.1. So I included a warning block that the user needs to specify the hydromt version in order to get the latest hydromt_wflow version. With the idea  that the warning block can be removed when hydromt_wflow is compatible with hydromt v1 (#351 #332). What do you think?

I have restricted the hydromt version now. Or would it make more sense to specify the hydromt_wflow version instead of hydromt?

### DIFF
--- a/docs/getting_started/installation.rst
+++ b/docs/getting_started/installation.rst
@@ -110,15 +110,28 @@ If you did not activate the shell you can still run the script in the environmen
   (hydromt-wflow) $ pixi run path/to/script.py
 
 If you intend to only use ``hydromt_wflow`` via the command line interface (CLI, see also the explanation in
-the `HydroMT-core docs <https://deltares.github.io/hydromt/stable/guides/user_guide/hydromt_cli.html>`_), then you can also install it globally using pixi like so:
+the `HydroMT-core docs <https://deltares.github.io/hydromt/stable/guides/user_guide/hydromt_cli.html>`_), then you can also install it globally using pixi.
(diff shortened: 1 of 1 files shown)


savente93 (approved):
very nice Sebastiaan, thank you!

PR #391, opened 2025-04-18
Fix/1d model connection

### LINKED ISSUE #390: MultiLineStrings in setup_1dmodel_connection after basin clipping

### Reproducible Example

A river geom (1d model) is provided to connect Wflow results to other models. This river geom should consist only of LineStrings which is handled by the workflow:

```
        logger.info("Linking 1D river to wflow river")
        # 1. Derive the river edges / boundaries
        # merge multilinestrings in gdf_riv to linestrings
        riv1d = gdf_riv.explode(index_parts=True).reset_index(drop=True)
        # clip to basins
        riv1d = riv1d.clip(basin_mask)
```
In a case, I still got an error due to MultiLineStrings being inside riv1d:
```
>
    riv1d_edges = riv1d.geometry.apply(lambda x: Point(x.coords[0]))
                                                       ^^^^^^^^
  File "C:\Users\hartgrin\.pixi\envs\hydromt\Lib\site-packages\shapely\geometry\base.py", line 1033, in coords
    raise NotImplementedError(
```
This is caused by clipping of riv1d to the basin:

Image

### Current behaviour

An error is thrown related to MultiLineStrings

### Desired behaviour

Give a clear error that shows the user that the river geometry is crossing model boundaries after clipping, and that the user needs to change the river geometry

### Additional context

_No response_

shartgring:
## Issue addressed
Fixes #390

## Explanation
Check for MultiLineString and if so, raise ValueError which guides the user in the right way

## Checklist
- [x] Updated tests or added new tests <- not needed
- [x] Updated documentation if needed <- not needed

None

### DIFF
--- a/hydromt_wflow/workflows/connect.py
+++ b/hydromt_wflow/workflows/connect.py
@@ -96,8 +96,9 @@ def wflow_1dmodel_connection(
     # Derive flwdir
     flwdir = hydromt.flw.flwdir_from_da(ds_model["flwdir"])
     # Basin mask
+    buffer_cells = 2
     basin_mask = ds_model["basins"].raster.vectorize()
-    basin_mask = basin_mask.buffer(-2 * max(ds_model.raster.res, key=abs))
+    basin_mask = basin_mask.buffer(-{buffer_cells} * max(ds_model.raster.res, key=abs))
 
     # If tributaries or subbasins area method,
     # need to derive the tributaries areas first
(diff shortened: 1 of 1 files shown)


shartgring:
I also notice `basin_mask = basin_mask.buffer(-2 * max(ds_model.raster.res, key=abs))` so I updated the error message

hboisgon (approved):
Thanks for improving the logging @shartgring! 
The issue is not solved though so maybe we could leave it open?

If you have time, I wonder if you could still try converting the multiline string into a singe line string by fetching the first edge node and the last edge node? If you don't have time could you share maybe the river geom file, and the config for `setup_basemaps` and `setup_rivers` to the linked issue so that we could reproduce later?

Thanks :)

shartgring:
@hboisgon I have done it using all the points from the MultiLineString, will that also work? And indeed I agree that it might be good to leave the issue open. I have also changed its description to focus less on the missing log message and more on the issue of clipping by the basin

Issue #394, opened 2025-04-24
Incorrect snapping in setup_1dmodel_connection caused by overlap with tributaries


shartgring:
### HydroMT-Wflow version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT-Wflow.

### Reproducible Example

I was working with coupling Wflow to a hydrodynamic model. Wflow and the 1dmodel overlap for a large portion. Only in the ends of the geometry there is some overlap with the tributaries instead of the main course of the river. This results in incorrect derivation of the coupling points between Wflow and the other model

![Image](https://github.com/user-attachments/assets/30dff9dc-0f08-418d-baca-63e7c5c19879)

I solved it by manually editing the 1dmodel river geometry which I provide to the setup method, but it took a bit of time to figure out that the problem was caused by the overlap

### Current behaviour

setup_1dmodel_connection cannot handle geometry ends that overlap with tributaries

### Desired behaviour

Either improve the coupling (use strahler order?) or raise a warning message when it notices that some strange snapping is occurring and that the user should/could edit the geometry of the 1dmodel

shartgring:
The staticmaps of the model are a bit large to upload to github (> 200 mb)

PR #416, opened 2025-05-15
Improvements to clipping and snapping in setup_1dmodel_connection

### LINKED ISSUE #390: MultiLineStrings in setup_1dmodel_connection after basin clipping

shartgring:
## Issue addressed
Fixes #390 and #394

## Explanation
I improved clipping by the basin 

## Checklist
- [X] Branch is up to date with `main`
- [X] Tests & pre-commit hooks pass

Clipping was previously done using a buffer on a rasterized basin mask which gave some strange edges:

image

It is now done based on removing cells from the mask itself and also made an optional argument for the user. Result with also 2 buffer cells:

image

Also, river segments are merged where possible. Without this, it would treat many loose ends as rivers which ended up with inconsistent subbasins derived in the workflow (like all cells assigned to subbasin):

image

And with merging included (snapping tolerance of 10 meters):

image


shartgring:
Okay SonarQube thinks that some new tests might be nice to improve coverage. Maybe good to use the existing test cases with some values for the tolerance, and also introduce a test case where snapping is tested?

shartgring:
image

The first adjustments worked fine when the 1d river geometry was preprocessed. Steps such as removing small side streams and some bifurcations (see also this image). Sometimes the small tributaries got snapped to the main river (so the gauge would be on the main river), which would result in all upstream area being considered as an upstream inflow area to this 'tributary' with incorrect snapping.

I added projection of the provided 1d river to the Wflow river map, and adding an option to filter by stream order. These small tributaries are of stream order 1 so filtering by 2 or larger solves the issue. I also made 2 the default value because of this reason

Also, when something goes wrong the whole model area gets assigned as a subbasin. So, I added a check that raises an error when this happens. This also prevents writing of very large geojson files for these subbasins. As it will create a geojson with each cell of the grid as a separate geometry

shartgring:
@hboisgon I am wondering whether it would be possible to refactor some of the code. For example, to move preprocessing of the geom to its own function (it's quite a bit of code now), and make a distinction between a workflow for subbasin_area method and one for nodes.

Both workflows can get called from setup_1dmodel_connection, and the logic of the combinations (like include_river_boundaries) could be moved to the setup method instead of being inside the workflow

And also to check if the nodes workflow can benefit from the preprocessing steps that I updated in the subbasin_area method. I only tested it for the subbasin_area method as I only needed that one for my project ;)

hboisgon (approved):
Thanks @shartgring for the work you did on this function! While testing for a small dummy case in piave basin, I found that your implementation was not fully robust yet. So here is a recap:

- I kept your function for the basin buffer, it works very nicely :)
- I could not reproduce the error you had when the 1d river crosses the basin boundaries. Both for the vecht and piave case. So I kept the original function to split the MultiLineStrings as the resulting shapefile looks more like what it should (connecting endpoints of the split river).
- If the 1d river tributaries are not well connected then you indeed need to snap them else we will get several rivers instead of one. Your snapping function gave weird results for piave. I wonder if it's because I use a shape from OSM where the rivers are not in the right order? But any I switched to endpoints snapping instead of lines. This should be more robust.
- On the how to remove river edges that are snapped to the main, eg very tiny line for a tributary. So I really want to try to avoid snapping the whole sobek river to wflow. Which is why I work a lot with river edges and then flow directions/raster operations. With the same concept, I instead use max uparea to find the most downstream river edge. Using flow directions, I then check for each upstream edge that the downstream one should be the outlet. Else this means a tributray has been snapped to the main river and should be removed. The disadvantage is that for now this means only 1 continuous 1driver can be connected to wflow. But I think this function could be extended later if we want.
- I kept your check that the whole basin is not derived and added an extra one to check that all subbasins do contain river cells.

I tested for piave and all the vecht cases you sent and all goes well. Hopefully this hsould lead to a much more robust function now :)

Issue #393, opened 2025-04-24
Update the default names in staticmaps


hboisgon:
### Kind of request

Changing existing functionality

### Enhancement Description

The default names in staticmaps are for most l;egacy names from the wflow python code (like wflow_reservoirareas, wflow_rivers, G_TT...). I think now would be a good time to go through these names and update.

For example use more lowercase convention and "_" (eg ThetaS to theta_s). 
These issue should probably be picked up together with #333

shartgring:
@hboisgon we will discuss this later today but I put these comments here as a reminder for myself as I am preparing it now :)

- lake parameters: also use new descriptive wflow notation (lake_e -> lake_rating_curve_exponent)
- water demand parameters: also use new descriptive wflow notation for those h1, h2, ..., variables
- n_land: I think something like land_manning or land_manning_n, and make it consistent with river_manning and floodplain_manning
- soil parameters: also append soil to the beginning like soil_ksat_vertical? And change soilthickness into soil_thickness. Same for vegetation
- sediment: I think also adding sediment_ in front of them will help. That might also help users recognize a wflow_sediment model 
- abbreviations or single letter variables in general: f, cf, cf_soil, tt, theta_r, etc. --> I think that these would also benefit from the new Wflow descriptive names and a prefix of what component they are relating to
- q_river_instanteneous and q_land_instanteneous --> more conistent to use river_instanteneous_q and land_instanteneous_q?
- h_river and h_land -> these are now water depths in Wflow (which also make sense), is the h symbol still appropiate? H is often used for water level, and D for depth
- forcing: I think that we can leave precipitation, pet and temp as they are or write them out. In any case, adding 'atmosphere_' to it feels a bit too much. For things like melting thresholds (which are in the staticmaps) it can be nice to use snowfall_temperature_threshold and snowpack_temperature_threshold

I think that in general my first ideas come down to this:
- appending the wflow component to **all** parameters (soil_, glacier_, river_, land_,)
- using the new wflow notation, especially for cryptic variable names
- leave out parts of the wflow notation where no ambiguity is expected, like snowfall_temperature_threshold instead of atmosphere_air__snowfall_temperature_threshold and soil_temperature_coefficient instead of soil_surface_temperature__weight_coefficient. I am doubting what to do with cases for vegetation and canopy: if it makes sense to use canopy_gap_fraction instead of vegetation_canopy_gap_fraction, because then to we change vegetation_water__storage_capacity to canopy_storage_capacity?

[...]

PR #418, opened 2025-05-19
Rename land river waterbodies static map variables

Tjalling-dejong:
## Issue addressed
Fixes partly #393 for Land / River / Waterbodies

## Explanation

## Checklist


hydromt_wflow/naming.py
```diff
@@ -359,18 +359,18 @@
...
         "wflow_v1": "river__width",
-        "hydromt_name": "rivwth",
+        "hydromt_name": "river_width",
     },
     # lakes
     "LakeArea": {
         "wflow_v0": "lateral.river.lake.area",
         "wflow_v1": "lake_surface__area",
-        "hydromt_name": "LakeArea",
+        "hydromt_name": "lake_area",
```
hboisgon, line 373: For lakes and reservoirs I also want to change the hydromt name as well as the default staticmaps name (the hydromt name is used in update mode, if you already have lakes.geojson in your geoms and use it to update some of the parameters of your lakes rather than go back and update the globla data).

hydromt_wflow/wflow.py
```diff
@@ -978,7 +978,7 @@ def setup_lulcmaps(
         lulc_vars: Dict = {
             "landuse": None,
             "Kext": "vegetation_canopy__light-extinction_coefficient",
-            "N": "land_surface_water_flow__manning_n_parameter",
+            "land_manning_n": "land_surface_water_flow__manning_n_parameter",
```
hboisgon, line 981: Here also the csv files need to be updated as well as parameters_data.yml

examples/data/wflow_upgrade/sbm/wflow_sbm_v0x.toml
hboisgon, line 1: For testing purpose, I think better if the v0 toml keep the old names as for almost all users, their v0 model are using these rather than the new ones

examples/data/wflow_upgrade/sediment/wflow_sediment_v0x.toml
hboisgon, line 1: Same: keep the v0 default names

hboisgon (commented):
Sorry for the confusion on hydromt_name vs default_staticmap_name in naming.py. I made the updates.
I also undid your changes on the v0 testing toml files for the conversion as I think I prefer our tests to be as close to what users will experience when using the function.
For the rest a few typos here and there or missing changes. I'll let you look at my commit :)

If you agree with the changes then we can merge!

PR #408, opened 2025-05-06
drop support for pcraster

LuukBlom:
## Issue addressed
Fixes #388 

## Explanation
Removed all pcraster occurrences
Updated example that used pcraster mapstacking to use the `setup_grid_from_raster` method

## Checklist
- [x] Updated tests or added new tests
- [x] Branch is up to date with `main`
- [x] Tests & pre-commit hooks pass
- [x] Updated documentation if needed

PR #421, opened 2025-05-20
update staticmaps_to_mapstack example

hboisgon:
## Issue addressed
Late review of updated notebook in PR #408 

## Explanation
Sorry if it was not clear in the issue but the part that fully updates the staticmaps based on all tif files is not something that works. It because some of the variable have a third dimension (layer or time) and when you merge, instead of getting a unique variable (eg c) you then get (c_1, c_2, c_3 etc) so Wflow does not run.

Else all good! I mainly added a final write at the end of the examples to avoid people saying that the changes do not work after running the notebook :) (it happened before for the 1D connection example notebook)

## Checklist

### DIFF
--- a/examples/convert_staticmaps_to_mapstack.ipynb
+++ b/examples/convert_staticmaps_to_mapstack.ipynb
@@ -31,10 +31,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import xarray as xr\n",
     "import shutil\n",
     "from os.path import join, exists\n",
-    "from os import listdir, rename\n",
+    "from os import listdir\n",
     "import hydromt\n",
     "from hydromt_wflow import WflowModel"
    ]
@@ -63,8 +62,8 @@
    "outputs": [],
    "source": [
     "root = \"wflow_piave_subbasin\"\n",
-    "mod = WflowModel(root, mode=\"r\")\n",
-    "ds = mod.grid  # here the staticmaps netcdf is loaded\n",
+    "wflow = WflowModel(root, mode=\"r\")\n",
+    "ds = wflow.grid  # here the staticmaps netcdf is loaded\n",
     "print(ds)"
    ]
   },
@@ -81,7 +80,7 @@
    "id": "8",
    "metadata": {},
    "source": [
-    "The raster module provides many raster GIS methods throught the **raster** Dataset accessor. To write a Dataset to a mapstack one line with code is sufficient. We only need to provide the output folder in which all raster files are saved. The default output format is *GeoTIFF*, but this can be changed with the `driver` argument. To write to PCRaster map-files it is recommended to have PCRaster python installed."
+    "The raster module provides many raster GIS methods through the **raster** Dataset accessor. To write a Dataset into several raster files (eg one geotiff file per map), one line with code is sufficient using [raster.to_mapstack](https://deltares.github.io/hydromt/stable/_generated/hydromt.gis.Dataset.raster.to_mapstack.html). We only need to provide the output folder in which all raster files are saved. The default output format is *GeoTIFF*, but this can be changed with the `driver` argument. To write to PCRaster map-files, pcraster python library should be installed."
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


examples/convert_staticmaps_to_mapstack.ipynb
```diff
@@ -114,110 +118,69 @@
...
-    "* read the updated mapstack\n",
-    "* change the model root to write the updated model to a new directory\n",
-    "* update the staticmaps of the model\n",
-    "* write the model\n",
-    "\n",
-    "> **Note:** We do not read the forcing as it is faster to just copy the file instead of loading it into python and writing it back to netcdf.\n",
+    "### Update a specific map using `setup_grid_from_raster`\n",
     "\n",
-    "> **Note:** The staticgeoms might be changed because of manual changes in the wflow_river, lakes, reservoir or glacier staticmaps and are therefore not read here. To change these maps we recommend using the hydromt update method to keep the staticgeoms and maps aligned."
+    "Lets say you have updated (manually or not) and the `KsatVer.tif` static map and that you would like to include it in your model. For demonstration purpose, we will make a copy and name it `KsatVer_updated.tif`."
```
deltamarnix, line 123:
```suggestion
    "Let's say you have updated (manually or not) the `KsatVer.tif` static map and that you would like to include it in your model. For demonstration purpose, we will make a copy and name it `KsatVer_updated.tif`."
```

examples/convert_staticmaps_to_mapstack.ipynb
```diff
@@ -114,110 +118,69 @@
...
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "15",
+   "id": "13",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "# read the entire updated mapstack\n",
-    "ds_updated = hydromt.open_mfraster(join(updated_staticmaps, \"*.tif\"))\n",
+    "You will need to find the `wflow variable` that corresponds to the map you updated, which in the case of `KsatVer.tif` is `soil_surface_water__vertical_saturated_hydraulic_conductivity`. For information on the available wflow variables, see the [Wflow documentation](https://deltares.github.io/Wflow.jl/dev/model_docs/parameters_intro.html)\n",
```
deltamarnix, line 143:
```suggestion
    "You will need to find the `wflow variable` that corresponds to the map you updated, which in the case of `KsatVer.tif` is `soil_surface_water__vertical_saturated_hydraulic_conductivity`. For information on the available wflow variables, see the [Wflow documentation](https://deltares.github.io/Wflow.jl/stable/model_docs/parameters_intro.html)\n",
```

I know the page doesn't exist yet, but I would recommend linking to stable, because that is eventually the release that will be sent out and people will want to use that documentation.

examples/convert_staticmaps_to_mapstack.ipynb
```diff
@@ -81,7 +80,7 @@
    "id": "8",
    "metadata": {},
    "source": [
-    "The raster module provides many raster GIS methods throught the **raster** Dataset accessor. To write a Dataset to a mapstack one line with code is sufficient. We only need to provide the output folder in which all raster files are saved. The default output format is *GeoTIFF*, but this can be changed with the `driver` argument. To write to PCRaster map-files it is recommended to have PCRaster python installed."
+    "The raster module provides many raster GIS methods through the **raster** Dataset accessor. To write a Dataset into several raster files (eg one geotiff file per map), one line with code is sufficient using [raster.to_mapstack](https://deltares.github.io/hydromt/stable/_generated/hydromt.gis.Dataset.raster.to_mapstack.html). We only need to provide the output folder in which all raster files are saved. The default output format is *GeoTIFF*, but this can be changed with the `driver` argument. To write to PCRaster map-files, pcraster python library should be installed."
```
LuukBlom, line 83:
```suggestion
    "The raster module provides many raster GIS methods through the **raster** Dataset accessor. To write a Dataset into several raster files (eg one geotiff file per map), one line with code is sufficient using [raster.to_mapstack](https://deltares.github.io/hydromt/stable/_generated/hydromt.gis.Dataset.raster.to_mapstack.html). We only need to provide the output folder in which all raster files are saved. The default output format is *GeoTIFF*, but this can be changed with the `driver` argument."
```
Missed this in my PR as well, but since we are dropping support for pcraster, should we delete the last line as well?

LuukBlom (approved):
LGTM! There are some suggestions that are up to you @hboisgon

examples/convert_staticmaps_to_mapstack.ipynb
hboisgon: Yes we can remove

PR #426, opened 2025-05-20
Small conversion fix

dalmijn:
## Issue addressed
Fixes error on absence of output header

## Explanation
Simply add the dictionary for 'output' when none is present. This is done under the if statements of netcdf (scalar) and csv. No test has been written because it would require a separate file, which seems overkill. The reason for the necessity of a separate file stems from the fact that the model is read/ re-read in `upgrade_to_v1_wflow`, which means that any pop operation on the config file is nullified.

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -32,6 +32,7 @@ Changed
 
 Fixed
 -----
+- Error when not output header is present in the v0.x toml to be converted
 - Updated installation guide (#376)
 
 Deprecated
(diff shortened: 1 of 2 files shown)


docs/changelog.rst
```diff
@@ -32,6 +32,7 @@ Changed
 
 Fixed
 -----
+- Error when not output header is present in the v0.x toml to be converted
```
hboisgon, line 35: You can remove this line as this feature is not yet released :)

hboisgon (approved):
Nice catch! You can remove the changelog line as the function has not been released yet so no need to mention a fix already

PR #431, opened 2025-05-27
Fix v1 attributes/ files

dalmijn:
## Issue addressed
Fixes stuff

## Explanation
Parameter data catalog was still in the v0.x format for HydroMT-core. So that has been ported to v1. Added an entry point for the wflow sediment model, as that was missing and causes an error.

## Checklist

No.


dalmijn:
Has been fixed

PR #432, opened 2025-05-27
feat: add geoms component

### LINKED ISSUE #400: Use GeomsComponent in v1
Use the geoms component from Hydromt core v1

Create `WflowGeomsComponent`:

- [ ] Instantiate but should completely inherit from core but docstrings can be udpated
- [ ] Initialise it in wflow.py and wflow_sediment.py init
- [ ] Test GeomsComponent? (should be tested mostly in core)

LuukBlom:
## Issue addressed
Fixes #400 

## Explanation
Implemented WflowGeomsComponent and updated old code to use it.

## Checklist


hydromt_wflow/wflow.py
```diff
@@ -82,7 +84,18 @@ def __init__(
         # This is when config_fn should be able to be passed to ConfigComponent later
         config_component = WflowConfigComponent(self, filename=str(config_path))
         grid_component = WflowGridComponent(self, filename=str(config_path))
-        components = {"config": config_component, "grid": grid_component}
+
+        geoms_component = WflowGeomsComponent(
+            model=self,
+            region_component="grid",  # TODO change when GridComponent is implemented? or set to None?  # noqa: E501
+            region_filename="geoms/geoms_region.geojson",  # TODO read from config?
+        )
```
LuukBlom, line 92: @savente93 @dalmijn, do you know how these `region_component` and `region_filename` are used? 

hydromt.model.components.GeomsComponent:
>         region_component: str, optional
>             The name of the region component to use as reference for this component's
>             region. If None, the region will be set to the union of all geometries in
>             the data dictionary.
>         region_filename: str
>             The path to use for writing the region data to a file. By default
>             "geoms/geoms_region.geojson".

What do I set `region_component` to?

Also, do you know if there is a config setting for the geoms output file?

savente93: So I think this is to do with the spacial component bit, Spacial components can do one of two things: 
1. define their own region (in this case by setting region_filename I think)
2. derive their region from another component like the grid (which is what the region_component is for) 
I think that if you inheret from geomscomponent in core (or spacialComponent if necessary) then you just have to set these to the correct thing and it _should_ just work from there. You can look in the GeomsComponent in core if you want to see it in action a bit more. I think that it makes sense to take our region from the grid, since everything is centred around that, but I'm happy to be corrected. Perhaps Brendan can shed some light there. 

Hope this helps, let me know if you need any more help

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,143 @@
...
+            raise ValueError(
+                f"hydrography_fn {hydrography_fn} not found in data catalog."
+            )
+        # Check on resolution (degree vs meter) depending on ds_org res/crs
+        scale_ratio = int(np.round(resolution / ds_org.raster.res[0]))
+        if scale_ratio < 1:
+            raise ValueError(
+                f"The model resolution {resolution} should be \
+larger than the {hydrography_fn} resolution {ds_org.raster.res[0]}"
+            )
```
Tjalling-dejong, line 61: I do not really understand this code. If the model resolution is for example 1/140 and the resolution of the dataset is 1/120 the error will still not be raised because of the np.round and casting it to int. Only if the model resolution is twice as small or smaller the error is raised.  i.e. np.round((1/240 ) / (1/120)) -> int( 0.5)  -> 0.

dalmijn (changes requested):
Before also reviewing the entire thing.
For compatibility and ease of adoptability I'd suggest renaming the WflowGeomsComponent to StaticGeomsComponent.

hydromt_wflow/components/geoms.py
hboisgon: This is a little linked to what happens in the hydrography workflow. We use scale_ratio to upscale the flow directions (>1 and do nothing ie keep original resolution if < 1).
The comment is maybe not the best but at this level the check is more meant to see if users area using the correct unit for the resolution ie the dem data is in degree (0.0008333) and they asked for a resolution in meter (100m) and vice versa. As there is no foolproof checks for this, these two were defined here. Maybe the value error message could be updated? I also do not really want the error if users ask for 90m and the dem data is 100m (they will get 100m but we dont have to throw an error). Maybe a warning then if scale_ratio=1 but resolutions are not equal?

hboisgon: Maybe another question seeing that part of the setup_basemaps was moved here (not sure why?) but do we really need a ``WflowGeomsComponent``? Can't we use core ``GeomsComponent`` directly or update it if we miss anything?

LuukBlom:
What is the standard we decided on? Wflow<ComponentName> right?

dalmijn:
> What is the standard we decided on? `Wflow<ComponentName>` right?

Yes

hydromt_wflow/components/geoms.py
hboisgon: Check for why we implement the Component.
For moving parts of the setup functions, we will do that in a second stage and go through setup methods one by one. Functionnality within setup methods actually belong more to (GIS) workflows (processes in v1 vocabulary) than components. For example the function you moved to Geoms in core moved to processes/basin_mask.py and processes/region.py and not in GeomsComponent.

hboisgon:
I forgot but please also update the api docs

hydromt_wflow/components/geoms.py
LuukBlom: Okay, I'll move the parse region stuff back into wflow.py then

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,157 @@
...
+        region_filename: str = "geoms/geoms_region.geojson",
+    ):
+        super().__init__(
+            model=model,
+            filename=filename,
+            region_component=region_component,
+            region_filename=region_filename,
+        )
+
+    def parse_region(
```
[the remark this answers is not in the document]
LuukBlom: Moved it to workflows.basemaps now!

hydromt_wflow/wflow_sediment.py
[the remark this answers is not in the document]
LuukBlom: WflowSedimentModel inherits from WflowModel and calls super().__init__() so that is actually already the case.

tests/test_components/test_geoms.py
```diff
@@ -0,0 +1,304 @@
...
+import xarray as xr
+from hydromt import DataCatalog
+from shapely.geometry import box
+
+from hydromt_wflow import WflowModel
+from hydromt_wflow.components import WflowGeomsComponent
+
+
+@pytest.fixture
+def mock_wflow_model():
```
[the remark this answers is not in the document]
LuukBlom: Missed this one apparently, but the models in conftest.py are actually real models created from the yml files. So these are used for the larger integration tests that are skipped for now. 

These mock models are not actually real models. no data goes in, all relevant data / function calls of the model are explicitly setup in the mock. This way its easy to test just this part in isolation, and be sure this functionality works as expected. 

When the components are all merged, we will of course start using the real models for testing as well.

hboisgon:
> > I forgot but please also update the api docs
> 
> What should I update specifically? I've looked through the docs and it doesn't seem to mention the geoms besides calling setup functions (which hasnt changed)



> > I forgot but please also update the api docs
> 
> What should I update specifically? I've looked through the docs and it doesn't seem to mention the geoms besides calling setup functions (which hasnt changed)

In this page: https://github.com/Deltares/hydromt_wflow/blob/feat/migration-hydromt-core-v1/docs/api.rst
I thought it had been done for config but not yet. It think good if we add a section for the Wflow components in our API docs especially for the methods set/read/write methods and future hydromt_steps methods if any. For example set_geoms calls were already replaced by geoms.set.

Updating the API docs has not been done yet so we can also decide to have a separate PR for it. What do you prefer?

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,157 @@
...
+    geopandas GeoDataFrames.
+    """
+
+    def __init__(
+        self,
+        model: Model,
+        *,
+        filename: str = "geoms/{name}.geojson",
+        region_component: Optional[str] = None,
+        region_filename: str = "geoms/geoms_region.geojson",
```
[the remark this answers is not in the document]
hboisgon: Sorry for all the comments...

I had a bit of a think and ``geoms.region`` does point to ``staticmaps.region`` in hydromt core v1 struct so the issue if we choose region (2D grid bounds) or basin is more for the ``WflowStaticmapsComponent`` than for you. 

EIse, I think same as ``GridComponent``, the region_filename is not really useful and in the case of ``GridComponent`` you cannot even skip them. I checked core and it seems that for ``GeomsComponent`` they do not even do anything?

So in this case I would go for None. I will review core behavior and harmonise / allow None if we do not want to write the component.region to file. (especially for components for which region_component is not None it does not really make sense to write this at all, you just end up with duplicated files of the same data)

hboisgon (changes requested):
Thanks for the updates @LuukBlom ! Almost there, just a few small things and a review of how dir_input is handled in read/write geoms.

Sorry for the confusion with region/basins, we did not really refine the model.region property that much when we did the first v1 skeleton and we should have... In the end this is an issue for WflowStaticmapsComponent and not WflowGeomsComponent so you can ignore my rambles ^^'

hydromt_wflow/workflows/basemaps.py
```diff
@@ -356,3 +364,93 @@ def topography(
...
     return ds_out
+
+
+def parse_region(
+    data_catalog: DataCatalog,
+    region: dict,
+    hydrography_fn: str | xr.Dataset,
+    resolution: float | int = 1 / 120.0,
+    basin_index_fn: str | xr.Dataset | None = None,
+) -> tuple[gpd.GeoDataFrame, Optional[np.ndarray], xr.Dataset, float]:
```
deltamarnix, line 375: That's quite a return statement. Would it make sense to split the function into smaller functions returning more specific objects?

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,159 @@
+import glob
+import logging
+from pathlib import Path
+from typing import Optional
+
+import geopandas as gpd
+from hydromt.model import Model
+from hydromt.model.components import GeomsComponent
+
+logger = logging.getLogger(__name__)
```
deltamarnix, line 10:
```suggestion
logger = logging.getLogger(f"hydromt.{__name__}")
```

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,159 @@
...
+            region_component=region_component,
+            region_filename=region_filename,
+        )
+
+    def read(
+        self,
+        read_dir: Path,
+        merge_data: bool = False,
+        **kwargs: dict,
+    ):
```
deltamarnix, line 58: Since we defined the following interface in `hydromt.GeomsComponent`, I would expect that the signature is the same for this inherited class. 

`def read(self, filename: Optional[str] = None, **kwargs) -> None:`

Would it be possible to stick to the signature and also make use of the `super.read()`?

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,159 @@
...
+    def get(self, name: str) -> Optional[gpd.GeoDataFrame | gpd.GeoSeries]:
+        """Get geometry by name."""
+        geom = self.data.get(name, None)
+        if geom is None:
+            logger.warning(f"Geometry '{name}' not found in geoms.")
+        return geom
+
+    def pop(
+        self, name: str, default=None
+    ) -> Optional[gpd.GeoDataFrame | gpd.GeoSeries]:
```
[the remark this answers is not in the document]
deltamarnix: The function actually doesn't do anything. A user could just as well use `geoms.data.pop` directly.

LuukBlom:
> In this page: https://github.com/Deltares/hydromt_wflow/blob/feat/migration-hydromt-core-v1/docs/api.rst I thought it had been done for config but not yet. It think good if we add a section for the Wflow components in our API docs especially for the methods set/read/write methods and future hydromt_steps methods if any. For example set_geoms calls were already replaced by geoms.set.
> 
> Updating the API docs has not been done yet so we can also decide to have a separate PR for it. What do you prefer?

I like doing the changes incrementally! Less likely to forget things + we avoid having a documentation sprint which noone likes doing. 

On another note though, users should not need to know about the components right? That is just a code architecture thing to keep it manageable for developers. In my opinion we should only document the api and what the users will use in their own scripts. That way, users wont start using functionality that we didn't intend for them to use, and then it becomes difficult to change.

hydromt_wflow/components/geoms.py
LuukBlom: Yes, good point, I'll make them equal

hydromt_wflow/wflow.py
```diff
@@ -4099,14 +4040,15 @@ def setup_1dmodel_connection(
...
+
         # Subcatchment map for river cells only (to be able to save river outputs
         # in wflow)
         self.set_grid(ds_out["subcatch_riv"], name=f"subcatchment_riv_{mapname}")
         gdf_subcatch_riv = ds_out["subcatch_riv"].raster.vectorize()
         gdf_subcatch_riv["value"] = gdf_subcatch_riv["value"].astype(
             ds_out["subcatch"].dtype
         )
-        self.set_geoms(gdf_subcatch_riv, name=f"subcatchment_riv_{mapname}")
+        self.set_geoms(gdf_subcatch_riv, name=f"subcatch_riv_{mapname}")
```
LuukBlom, line 4051: @hboisgon is this also a merge conflict I messed up?

I cant find it in naming.py

hydromt_wflow/wflow.py
```diff
@@ -5330,36 +5272,34 @@ def read_geoms(
...
+            read_dir=read_dir,
+            merge_data=not self._write,
+        )
 
     @hydromt_step
     def write_geoms(
         self,
         geoms_fn: str = "staticgeoms",
         precision: int | None = None,
+        to_wgs84: bool = False,
```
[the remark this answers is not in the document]
hboisgon: I agree we do not have to document everything about the components but I also don't think we need to hide them from users. Else we will need a lot of duplicates of functionnality in the ``WflowModel``. I already mentioned ``set`` but if the users just want to access the direct python object e.g. for plotting, currently this is through ``WflowModel.geoms.data``. Our config object is also complex (tomldocument) and we need custom methods to set, get, pop that the users should also be able to access. 

A lot of users were using the python objects ``geoms``, ``grid``, ``config`` and these fundamentally changed to classes and not dict or xarray objects, so we cannot hide this change to them or they will have surprises. I agree for CLI users it won't make a difference but for python users it will. (and it matters to plugin developers too). Maybe we could split the API doc into different pages, one for CLI users and one for python users?

It;s a change that came from core but it's the same for ModelRoot: it's now a class and not a path that is accessed via model.root.path. Also useful property e.g. for saving your plots in a folder relative to your Wflow model directory. Do we then also duplicate a shortcut to it in WflowModel?

hydromt_wflow/wflow.py
hboisgon: Yes it is. It's not in naming.py, it's not a wflow variable but the name of a map used to save specific outputs for specific subbasins. We did rename the wflow variable from *wflow_subcatch* to *subcatchment* so I harmonised for this output as well. You can just check what is in main in these cases too if you hesitate during conflicts :)

hydromt_wflow/wflow.py
```diff
@@ -5330,36 +5284,73 @@ def read_geoms(
...
 
-        fns = glob.glob(join(dir_mod, "*.geojson"))
-        if len(fns) > 1:
-            logger.info("Reading model staticgeom files.")
-        for fn in fns:
-            name = basename(fn).split(".")[0]
-            if name != "region":
-                self.set_geoms(gpd.read_file(fn), name=name)
+        if not self._write:
+            self.geoms.clear()  # start fresh in read-only mode
```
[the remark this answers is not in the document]
LuukBlom: This is the behaviour that is in main though.

https://github.com/Deltares/hydromt_wflow/blob/b8995b393cd1a856a6c43ac58bb6aa25247c773d/hydromt_wflow/wflow.py#L5284

Do you have a suggestion for how to handle it?

hydromt_wflow/components/geoms.py
```diff
@@ -0,0 +1,159 @@
...
+            Additional keyword arguments that are passed to the
+            `geopandas.read_file` function.
+        """
+        fns = glob.glob(filename)
+        if fns:
+            logger.info("Reading model staticgeom files.")
+        for fn in fns:
+            name = Path(fn).stem
+            if name != "region":
+                self.set(gpd.read_file(fn, **kwargs), name=name)
```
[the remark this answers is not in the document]
LuukBlom: I wasnt using core because of the region being excluded from setting, but in another comment you mentioned that is no longer needed so I will use core

tests/conftest.py
```diff
@@ -211,3 +212,22 @@ def demda():
     # NOTE epsg 3785 is deprecated https://epsg.io/3785
     da.raster.set_crs(3857)
     return da
+
+
+@pytest.fixture
+def mock_rasterdataset(mocker: MockerFixture) -> xr.Dataset:
```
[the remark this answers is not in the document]
LuukBlom: It is used in `test_workflows/test_basemaps.py` and in `test_components/test_geoms_component.py`.

Are you saying you want to have 2 copies of this fixture? One in the conftest.py for `test_workflows` and one in `test_components`?

tests/test_components/test_config_component.py
```diff
@@ -11,6 +11,12 @@
 from hydromt_wflow.utils import DATADIR
 
 
+@pytest.fixture
+def mock_model(mock_model_factory) -> MagicMock:
+    """Create a mock model with a root."""
+    return mock_model_factory(mode="w")
```
hboisgon, line 17: I just merged states so maybe you'll need to add it in test_states_component too ?

hboisgon (approved):
All good! Ready to merge now :)

PR #436, opened 2025-06-03
Config component

### LINKED ISSUE #342: Config component
### Kind of request

We need to add the ConfigComponent to the WflowModel.

Also, the config component needs to be able to write groups of variables into one.

For example

```toml
[input.vertical]
khorfrac.value = 100
kverfrac.value = 200

# instead of 

[input.vertical.khorfrac]
value = 100

[input.vertical.kverfract]
value = 200
```

Also, there are values that contain a tilde (~) and they need to be quoted ("")

```toml
vegetation_canopy_water__mean_evaporation-to-mean_precipitation_ratio.value = 0.11
"soil~compacted_surface_water__infiltration_capacity".value = 5
soil_water_sat-zone_bottom__max_leakage_volume_flux.value = 0
"soil_root~wet__sigmoid_function_shape_parameter".value = -500
atmosphere_air__snowfall_temperature_threshold.value = 0
atmosphere_air__snowfall_temperature_interval.value = 2
 ```

- [ ] Add WflowConfigComponent
- [x] Customize the write functionality to support the cases defined above. See #79 
- [ ] Add new tests
- [ ] Merge to v1 branch

### Enhancement Description

For the move to hydromt-core v1 we should use a config component. 

### Use case

_No response_

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by dalmijn**
## Issue addressed
Fixes #342

## Explanation
Created a custom wflow config component based on the base ModelComponent class. 
This class uses `tomlkit.TOMLDocument` as it's internal data structure.
`set_config` and `get_config` methods were taken from the general utils submodule.

## Checklist

No.

### DIFF
--- a/hydromt_wflow/__init__.py
+++ b/hydromt_wflow/__init__.py
@@ -1,10 +1,7 @@
 """hydroMT plugin for wflow models."""
 
-__version__ = "0.8.1.dev0"
-
 from .naming import *
 from .utils import *
-from .wflow import WflowModel
-from .wflow_sediment import WflowSedimentModel
-
-__all__ = ["WflowModel", "WflowSedimentModel"]
+from .version import __version__
+from .wflow import *
+from .wflow_sediment import *
--- a/hydromt_wflow/__init__.py
+++ b/hydromt_wflow/__init__.py
@@ -1,7 +1,8 @@
-"""hydroMT plugin for wflow models."""
+"""HydroMT plugin for wflow models."""
 
-from .naming import *
-from .utils import *
-from .version import __version__
-from .wflow import *
-from .wflow_sediment import *
+__version__ = "1.0.0.dev"
+
+from .wflow import WflowModel
+from .wflow_sediment import WflowSedimentModel
+
+__all__ = ["WflowModel", "WflowSedimentModel"]
--- a/hydromt_wflow/components/__init__.py
+++ b/hydromt_wflow/components/__init__.py
@@ -1,4 +1,6 @@
 """Wflow model components submodule."""
 
-from .config import *
-from .staticmaps import *
+from .config import WflowConfigComponent
+from .staticmaps import StaticmapsComponent
+
+__all__ = ["StaticmapsComponent", "WflowConfigComponent"]
--- a/hydromt_wflow/components/__init__.py
+++ b/hydromt_wflow/components/__init__.py
@@ -1,6 +1,4 @@
-"""submodule for model components."""
+"""Wflow model components submodule."""
 
-from hydromt_wflow.components.config import WflowConfigComponent
-from hydromt_wflow.components.grid import WflowGridComponent
-
-__all__ = ["WflowConfigComponent", "WflowGridComponent"]
+from .config import *
+from .staticmaps import *
--- a/hydromt_wflow/components/config.py
+++ b/hydromt_wflow/components/config.py
@@ -49,6 +49,11 @@ class WflowConfigComponent(ModelComponent):
 
     def __eq__(self, other: ModelComponent):
         """Compare components based on content."""
+        if not isinstance(other, WflowConfigComponent):
+            raise ValueError(
+                f"Can't compare {self.__class__.__name__} \
+with type {type(other).__name__}"
+            )
         other_config = cast(WflowConfigComponent, other)
         return self.data == other_config.data
 
--- a/hydromt_wflow/components/config.py
+++ b/hydromt_wflow/components/config.py
@@ -1,17 +1,24 @@
 """Custom wflow config component module."""
 
-from typing import Any
+import logging
+from pathlib import Path
+from typing import Any, cast
 
 import tomlkit
 from hydromt.model import Model
-from hydromt.model.components.config import ConfigComponent
+from hydromt.model.components.base import ModelComponent
+from hydromt.model.steps import hydromt_step
+from tomli_w import dump as dump_toml
+from tomllib import load as load_toml
 
 from hydromt_wflow.utils import get_config, set_config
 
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: tests/components/test_config_component.py [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+    # Setup the component
+    component = WflowConfigComponent(mock_model)
+
+    # Update it like a dummy to request
+    component.data.update(config_dummy_data)
+
+    # Assert asking for entry
+    assert component.get("biem") == "bam"
+    assert component.get("time") == {"sometime": "now"}
+    assert isinstance(component.get("foo"), dict)
```
savente93, line 39: This will pass because tomlkit does some shenanigans but it's not stricktly correct. I think you want to assert that it is an instance of a tomlkit AbstractTable (super class for tables and AoTs iirc)

tests/components/test_config_component.py
```diff
@@ -0,0 +1,181 @@
+import logging
+from pathlib import Path
+from unittest.mock import PropertyMock
+
+import pytest
+from hydromt.model import ModelRoot
```
savente93, line 6: Code is well tested and I like the unti tests you wrote, well done!

hydromt_wflow/version.py
```diff
@@ -0,0 +1,9 @@
+"""HydroMT-Wflow version."""
+
+main = 1
+major = 0
+minor = 0
+build = 0
```
savente93, line 6: please use the actual semver names which is `major.minor.patch` if you want to do it like this

hydromt_wflow/__init__.py
```diff
@@ -1,10 +1,7 @@
...
-
 from .naming import *
 from .utils import *
-from .wflow import WflowModel
-from .wflow_sediment import WflowSedimentModel
-
-__all__ = ["WflowModel", "WflowSedimentModel"]
+from .version import __version__
+from .wflow import *
+from .wflow_sediment import *
```
savente93, line 7: I'm not a fan of star imports, especially if we didn't set the `__all__` because this is very bad for type checkers. Can we use more direct imports here?

hydromt_wflow/components/__init__.py
```diff
@@ -1,6 +1,4 @@
-"""submodule for model components."""
+"""Wflow model components submodule."""
 
-from hydromt_wflow.components.config import WflowConfigComponent
-from hydromt_wflow.components.grid import WflowGridComponent
-
-__all__ = ["WflowConfigComponent", "WflowGridComponent"]
+from .config import *
+from .staticmaps import *
```
savente93, line 4: I'm not a fan of star imports, especially if we didn't set the `__all__` because this is very bad for type checkers. Can we use more direct imports here?

hydromt_wflow/components/config.py
```diff
@@ -1,43 +1,131 @@
...
         *,
         filename="wflow_sbm.toml",
-        default_template_filename: str | None = None,
     ):
-        super().__init__(
-            model=model,
-            filename=filename,
-            default_template_filename=default_template_filename,
-        )
+        self._data: tomlkit.TOMLDocument[str, Any] | None = None
```
savente93, line 45: I'm not totally sure if TOMLDocument is generic like that, you'd have to double check that. Otherwise it's just `tomlkit.TOMLDocument`

hydromt_wflow/components/config.py
```diff
@@ -1,43 +1,131 @@
...
-        set_config(self._data, *args)
+        super().__init__(model=model)
 
-    def get(self, *args) -> Any | None:
-        """Get config options."""
-        self._initialize()
-        return get_config(self._data, *args)
+    def __eq__(self, other: ModelComponent):
+        """Compare components based on content."""
+        other_config = cast(WflowConfigComponent, other)
```
savente93, line 52: Typically checking that `other` is of the same type of self is part of `__eq__` implementations. cast doesn't actually do anything except for type checkers, so I'd replace this with an 

```python
if not isinstance(other, WflowConfigCompoent):
    return False
```
or something similar

hydromt_wflow/components/config.py
```diff
@@ -1,43 +1,131 @@
...
+        """Read the wflow configurations file."""
+        self._initialize(skip_read=True)
+
+        # Solve pathing
+        p = path or self._filename
+        read_path = Path(self.root.path, p)
+
+        # Check if the file exists
+        if read_path.is_file():
+            logger.info(f"Reading model config file from {read_path.as_posix()}.")
```
savente93, line 86: I think we should follow the system convention that we are running on so I'd remove the `as_posix()` (even though I also prefer it)

tests/components/test_config_component.py
```diff
@@ -0,0 +1,181 @@
...
+    # Setup the component
+    component = WflowConfigComponent(mock_model)
+
+    # Assert that the internal data is None
+    assert component._data is None
+
+    # When asking for data property, it should return a tomlkit document
+    assert isinstance(component.data, TOMLDocument)
+    assert isinstance(component._data, TOMLDocument)  # Same for internal
+    assert len(component.data) == 0
```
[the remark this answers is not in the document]
dalmijn: For the moment, we don't use a default config?

tests/components/test_config_component.py
dalmijn: Fair enough, will change.

hydromt_wflow/__init__.py
dalmijn: Will revert.

tests/components/test_config_component.py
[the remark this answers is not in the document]
savente93: good point then we can keep it like that for now

PR #438, opened 2025-06-04
fix: docs gen by removing broken references

aslowwriter:
References broke either due to the removal of the attributes or because the examples stopped wroking. These refeerneces are remvoed now because commenting them out didn't seem to make a difference. We'll just have to remember to add these references back in when we renable the examples etc.
 
After this, the docs should be green again, and I think that people making PRs to create new components should add them to the documentation. We can fix up examples etc later when everthing is working again, but they should at least be included in the `docs/api.rst` 

## Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -286,5 +286,5 @@ nbsphinx_prolog = r"""
         </div>
 """
 
-nbsphinx_execute = "always"
+# nbsphinx_execute = "always"
 nbsphinx_timeout = 300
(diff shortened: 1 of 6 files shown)


dalmijn (changes requested):
Thanks for the good update Sam!

Could you comment the examples notebooks out instead of removing them?

dalmijn (approved):
LGTM, thanks Sam!

PR #454, opened 2025-06-10
Feat: Staticmaps Component

### LINKED ISSUE #343: Use Grid component in v1
### Kind of request

### Enhancement Description

For the move to hydromt-core v1 we should use the grid component, instead of the GridModel from pre v1. 

- [ ] Create a WflowGridComponent
- [ ] Migrate write and read functions from WflowModel
- [ ] Create tests
- [ ] Merge to v1 branch

### Use case

_No response_

### Additional Context

_No response_

dalmijn:
## Issue addressed
Fixes #343 

## Explanation
Inherited from `GridComponent` from core. The `read` method was overwritten to incorporate config information and the ability to flip the data upside down. The `write` method was overwritten to make use of the config information and to set information back into the config, it does however supercharge. Also the `write` method now makes paths relative to the config parent location. For this purpose the functions from core were rewritten. The `set` method was overwritten to incorporate the functionality in regards to cyclic maps and masking of data. Added a method to update the naming of the data variables in the dataset.

## Checklist

No.

### DIFF
--- a/hydromt_wflow/components/utils.py
+++ b/hydromt_wflow/components/utils.py
@@ -14,7 +14,10 @@ MOUNT_PATTERN = re.compile(r"(^\/(\w+)\/|^(\w+):\/).*$")
 def _mount(
     value: str,
 ) -> str | None:
-    """Get the mount of a path."""
+    """Get the mount of a path.
+
+    As this not properly solved currently by pathlib or os.
+    """
     m = MOUNT_PATTERN.match(value)
     if m is None:
         return None
@@ -25,7 +28,12 @@ def _relpath(
     value: Any,
     root: Path,
 ) -> str | Any:
-    """Generate a relative path."""
+    """Generate a relative path.
+
+    Being able to go either up or down in directories.
+    Also return the original path if the mount differs.
+    Otherwise it will error..
+    """
     if not isinstance(value, (Path, str)) or not Path(value).is_absolute():
         return value
     value = Path(value)


hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+                self._data = self.data.drop_vars(vars_to_drop + ["layer"])
 
-__all__ = ["StaticmapsComponent"]
+        # Determine the masking layer
+        mask = get_mask_layer(mask, self.data, data)
 
+        # Set the data per layer
+        for dvar in data.data_vars:
+            # TODO removed reading mode check, review the exact effect of this
+            # I personally couldn't see the point..
```
hboisgon, line 219: Indeed
```suggestion
```

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+            "input.path_static",
+            Path(self.root.path, p_input).as_posix(),
+        )
+
+    ## Mutating methods
+    def set(
+        self,
+        data: xr.DataArray | xr.Dataset,
+        name: str | None = None,
+        mask: str | xr.DataArray | None = None,
```
hboisgon, line 149: I don't think this needs to be an argument? It's only temporay that we still have model.set_grid to avoir merge conflicts but it will be removed at some point (especially since it is not even an hydromt_step) so you can fully transfer now already.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+                logger.warning(
+                    f"Replacing 'layer' coordinate, dropping variables \
+({vars_to_drop}) associated with old coordinate"
+                )
+                # Use `_grid` as `grid` cannot be set
+                self._data = self.data.drop_vars(vars_to_drop + ["layer"])
 
-__all__ = ["StaticmapsComponent"]
+        # Determine the masking layer
+        mask = get_mask_layer(mask, self.data, data)
```
hboisgon, line 214: Is this really worth it to create this one use function with multiple args? In our case mask layer is quite fixed. I had imagined the function I had in utils could be re-used in processes or workflows as it was quite generic but for this one I wonder?

```suggestion
        mask_name = self.model._MAPS["basins"]
        if mask_name in self.data:
          mask = self.data[mask_name] != self.data[mask_name].raster.nodata
        elif mask_name in data:
          mask = data[mask_name] != data[mask_name].raster.nodata
        else:
          mask = None
```

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+                    data[dvar] = data[dvar].where(mask, False)
+            self._data[dvar] = data[dvar]
 
-class StaticmapsComponent(GridComponent):
-    """wflow staticmaps component."""
+    ## Setup and update methods
+    @hydromt_step
+    def update_names(
+        self,
+        **select,
```
hboisgon, line 236:
```suggestion
        **rename,
```

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+    ):
+        """Map the names of the data variables to new ones.
```
hboisgon, line 239:
```suggestion
        """Map the names of the data variables to new ones.
        
        This method however does not change the new names in the TOML file. 
        To update them, you can use it together with the `WflowModel.setup_config` method.

```

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
 
-    pass
+        Parameters
+        ----------
+        select : dict, optional
```
hboisgon, line 242: It's internal but in the docstrings still so I harmonised with names in other core methods

```suggestion
        rename : dict, optional
```

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+        """
+        # Check whether they are in the maps
+        nf = []
+        for key in list(select.keys()):
+            if key not in self.data.data_vars:
+                nf.append(key)
+                _ = select.pop(key)
+        if len(nf) != 0:
+            logger.warning(f"Could not rename {nf}, not found in data")
+        self._data = self.data.rename(**select)
```
hboisgon, line 254:
```suggestion
        not_found = []
        for key in list(rename.keys()):
            if key not in self.data.data_vars:
                not_found.append(key)
                _ = rename.pop(key)
        if len(not_found) != 0:
            logger.warning(f"Could not rename {not_found}, not found in data")
        self._data = self.data.rename(**rename)
```

hydromt_wflow/components/config.py
```diff
@@ -129,8 +130,11 @@ def write(
             logger.info(f"Writing model config to {write_path.as_posix()}.")
             write_path.parent.mkdir(parents=True, exist_ok=True)
 
+            # Solve the pathing in the data
+            write_data = make_config_paths_relative(self.data, write_path.parent)
```
hboisgon, line 134: Do we really need this? Quid of thes cases where dir_input exists?

hydromt_wflow/wflow.py
```diff
@@ -5036,43 +5033,6 @@ def write_config(
         # Call the component
         self.config.write(p)
 
-    @hydromt_step
-    def read_grid(self, **kwargs):
```
hboisgon, line 5040: You deleted the read but kept the write and set? I was not there when the decision was made but good to be consistent

hydromt_wflow/wflow.py
```diff
@@ -5173,51 +5088,8 @@ def set_grid(
...
-            if name is not None:
-                logger.warning(f"Layer {name} will not be masked with basins.")
-        elif self._MAPS["basins"] in self.grid:
-            data = mask_raster_from_layer(data, self.grid[self._MAPS["basins"]])
-        elif self._MAPS["basins"] in data:
-            data = mask_raster_from_layer(data, data[self._MAPS["basins"]])
-        # fall back on default set_grid behaviour
-        super().set_grid(self, data, name)
+        # Call the staticmaps set method
+        self.staticmaps.set(data, name=name, mask=self._MAPS["basins"])
```
hboisgon, line 5092: This method will disappear from model so good to already have mask taken care of in the component

hboisgon (changes requested):
Nice implementation! Just a few things here and there to finalize.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+        # Set the data per layer
+        for dvar in data.data_vars:
+            # TODO removed reading mode check, review the exact effect of this
+            # I personally couldn't see the point..
+            if dvar in self._data:
+                logger.warning(f"Replacing grid map: {dvar}")
+            if mask is not None:
+                # TODO maybe we could outlaw boolean layers
+                # Instead use 0 and 1 for actual data and -1 or -9999 for all i care
+                # for the nodata value
```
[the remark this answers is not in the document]
dalmijn: Boolean data is essentially 0 and 1's. Dont think wflow.jl will care that much. But will talk to Willem about that.

hydromt_wflow/wflow.py
hboisgon, line 1: I forgot but you also should replace all calls to `self.grid` into `self.staticmaps.data` in wflow.py

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+        By default "staticmaps.nc".
+    region_component : str, optional
+        The name of the region component to use as reference
+        for this component's region. If None, the region will be set to the grid extent.
+        Note that the create method only works if the region_component is None.
+        For add_data_from_* methods, the other region_component should be
+        a reference to another grid component for correct reprojection, by default None
+    region_filename : str
+        The path to use for reading and writing of the region data by default.
+        By default "region.geojson".
```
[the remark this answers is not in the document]
dalmijn: I'll adjust it then.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+
+    def __init__(
+        self,
+        model: Model,
+        *,
+        filename: str = "staticmaps.nc",
+        region_component: str | None = None,
+        region_filename: str = "region.geojson",
+    ):
+        GridComponent.__init__(
```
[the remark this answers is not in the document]
dalmijn: Yep

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+    def read(
+        self,
+        filename: Path | str | None = None,
+        *,
+        mask_and_scale: bool = False,
+        **kwargs,
+    ):
+        """Read staticmaps model data.
+
+        Key-word arguments are passed to :py:meth:`~hydromt.model.Model.read_nc`
```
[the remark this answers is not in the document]
dalmijn: Alright.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+            _FillValue, scale_factor and add_offset are taken from variable attributes
+            (if they exist).
+        **kwargs : dict
+            Additional keyword arguments to be passed to the `read_nc` method.
+        """
+        self.root._assert_read_mode()
+        self._initialize_grid(skip_read=True)
+
+        # Sort which path/ filename is actually the one used
+        # Hierarchy is: 1: signature, 2: config, 3: default
```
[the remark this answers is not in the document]
dalmijn: Signature is more clear to me from a dev pov

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+
+        Parameters
+        ----------
+        filename : str, optional
+            filename relative to model root, by default None
+        **kwargs : dict
+            Additional keyword arguments to be passed to the `write_nc` method.
+        """
+        # Solve pathing same as read
+        # Hierarchy is: 1: signature, 2: config, 3: default
```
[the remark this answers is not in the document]
dalmijn: Again, signature is more clear to me.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+    @hydromt_step
+    def read(
+        self,
+        filename: Path | str | None = None,
+        *,
+        mask_and_scale: bool = False,
+        **kwargs,
+    ):
+        """Read staticmaps model data.
+
```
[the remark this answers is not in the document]
dalmijn: Fair.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+            self.set(ds)
+
+    @hydromt_step
+    def write(
+        self,
+        filename: Path | str | None = None,
+        **kwargs,
+    ):
+        """Write staticmaps model data.
+
```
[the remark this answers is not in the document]
dalmijn: Also fair.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+        Key-word arguments are passed to :py:meth:`~hydromt.model.Model.write_nc`
```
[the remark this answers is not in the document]
dalmijn: Alright.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+        p_input = Path(self.model.config.get("dir_input", fallback=""), p)
+
+        # Supercharge with the base grid component write method
+        super().write(
+            p_input.as_posix(),
+            gdal_compliant=True,
+            rename_dims=True,
+            **kwargs,
+        )
+
```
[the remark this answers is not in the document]
dalmijn: I know, but the second part I checked whether that is still an issue with wflow.jl, it is not. The first part is covered by gdal_compliant.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+    ## Mutating methods
+    def set(
+        self,
+        data: xr.DataArray | xr.Dataset,
+        name: str | None = None,
+        mask: str | xr.DataArray | None = None,
+    ):
+        """Add data to the staticmaps.
+
+        All layers of grid must have identical spatial coordinates.
```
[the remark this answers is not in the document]
dalmijn: Not a big fan of these types of large docstrings, but if you're really in favour of having this one then I'm fine with it.

hydromt_wflow/components/staticmaps.py
dalmijn: fair.

hydromt_wflow/components/staticmaps.py
dalmijn: Sure.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+            Additional keyword arguments to be passed to the `read_nc` method.
+        """
+        self.root._assert_read_mode()
+        self._initialize_grid(skip_read=True)
+
+        # Sort which path/ filename is actually the one used
+        # Hierarchy is: 1: signature, 2: config, 3: default
+        p = filename or self.model.config.get("input.path_static") or self._filename
+        # Check for input dir
+        p_input = self.model.config.get("dir_input", fallback="")
```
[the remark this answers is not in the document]
dalmijn: I think you mean `super().read()`? `flipud` is called during settings of the datasets, which is done per dataset.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,254 @@
...
+            gdal_compliant=True,
+            rename_dims=True,
+            **kwargs,
+        )
+
+        # Set the config entry to the correct path
+        self.model.config.set(
+            "input.path_static",
+            Path(self.root.path, p_input).as_posix(),
+        )
```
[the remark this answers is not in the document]
dalmijn: I think settings the full paths in the config and solving that when writing the config is the safest. Tracking 'how relative' paths are is quite hazardous.

hydromt_wflow/components/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    ----------
+    data : TOMLDocument
+        The configurations in a TOMLDocument format.
+    root : Path
+        The root to which the paths are made relative.
+        Most of the time, this will be the parent directory of the
+        configurations file.
+    """
+    for key, val in data.items():
+        if isinstance(val, Table):
```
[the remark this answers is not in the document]
savente93: If you want you can leave it as is.

hydromt_wflow/components/staticmaps.py
hboisgon: Ok. Then I would still for now maybe remove the mask argument and simplify to this:

```suggestion
        mask = get_mask_layer(self.model._MAPS["basins"], self.data, data)
```

hydromt_wflow/components/staticmaps.py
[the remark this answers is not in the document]
hboisgon: I think you opened the issue so you can remove the comment and leave as is for now?

hydromt_wflow/components/staticmaps.py
hboisgon: I see you did it below but forgot it in the docstrings :)

hboisgon (commented):
Almost there :) some small stuff here and there

dalmijn:
@hboisgon: "I think it's not very clean to have a FillValue fow x_dim and y_dim. I think Wflow.jl actually wrote some workaround code while they were waiting for hydromt_wflow to solve it so it works but it's not the best way I think (I checked and it's actually against CF conventions to have a FillValue for coordinates). You could argue we could take care of this in core too with a new argument like force_sn. What do you think?"

I feel like this should be taken care of in core. Indeed, it's not in accordance with the cf conventions

hydromt_wflow/wflow.py
hboisgon: Should be back in (read_staticmaps)

hydromt_wflow/wflow.py
```diff
@@ -5139,43 +5153,6 @@ def write_config(
...
-            ds = xr.load_dataset(
-                fn, mask_and_scale=False, decode_coords="all", **kwargs
-            )
-            # make sure internally maps are always North -> South oriented
-            if ds.raster.res[1] > 0:
-                ds = ds.raster.flipud()
-            self.set_grid(ds)
-
     @hydromt_step
     def write_grid(
```
hboisgon, line 5157:
```suggestion
    def write_staticmaps(
```

hydromt_wflow/wflow.py
```diff
@@ -5195,58 +5172,13 @@ def write_grid(
...
-        if not isdir(dirname(fn)):
-            os.makedirs(dirname(fn))
-        logger.info(f"Write grid to {fn}")
-
-        ds_out.to_netcdf(fn, encoding=encoding)
+        # Call the component write method
+        self.staticmaps.write(filename=fn_out)
 
     @hydromt_step
     def set_grid(
```
hboisgon, line 5179:
```suggestion
    def set_staticmaps(
```

hboisgon (commented):
Almost there! Also update the API docs and changelog

hydromt_wflow/components/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    if not isinstance(value, (Path, str)) or not Path(value).is_absolute():
+        return value
+    value = Path(value)
+    try:
+        if _mount(value.as_posix()) == _mount(root.as_posix()):
+            value = Path(relpath(value, root))
+    except ValueError:
+        pass  # `value` path is not relative to root
+    finally:
+        return value.as_posix()
```
deltamarnix, line 40: I think this function can be removed all together and instead we should use [`PurePath.relative_to()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to).

It gives a ValueError if it is not able to create a relative path.

Tested with:

```py
a = Path("/d/tmp/foo")
a.relative_to("/d/")
# returns 'tmp/foo'
```

hydromt_wflow/components/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def make_config_paths_relative(
+    data: TOMLDocument,
```
deltamarnix, line 44: TomlDocument no longer exists with Tjallings rollback. It is now a dictionary you walk through.

hydromt_wflow/components/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Parameters
+    ----------
+    mask : str | xr.DataArray | None
+        The mask itself or the name of the mask layer in another dataset.
+    *args : list
+        These have to be xarray Datasets in which the mask (as a string)
+        can be present
+    """
+    if mask is None:
+        return mask
```
deltamarnix, line 82: I think this is confusing. Maybe just not allow None?

Or otherwise maybe just write
```suggestion
    if mask is None:
        return None
```

hydromt_wflow/components/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    if isinstance(mask, xr.DataArray):
+        return mask != mask.raster.nodata
+    if not isinstance(mask, str):
+        raise ValueError(f"Unknown type for determining mask: {type(mask).__name__}")
```
deltamarnix, line 86: I always recommend error raising at the top of the function

deltamarnix (changes requested):
Two things that catch my eye:

- I think that the utils functions for relative paths can be removed, because pathlib already implements what you want.
- Instead of changing wflow.py all over the place, I expected that we would keep wflow as much as possible as we had it, and then later when we have everything in place we do that big renaming PR. But maybe staticmaps is a special one, because we are removing the word `grid` from the concept?

hydromt_wflow/components/utils.py
dalmijn: No.

hydromt_wflow/wflow.py
```diff
@@ -113,7 +113,81 @@ def __init__(
...
+    def states(self) -> WflowStatesComponent:
+        """Return the states component."""
+        return self.components["states"]
+
+    @property
+    def staticmaps(self) -> WflowStaticmapsComponent:
+        """Return the staticmaps component."""
+        return self.components["staticmaps"]
+
+    # Wflow specific
```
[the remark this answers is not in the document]
dalmijn: What?

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,239 @@
...
+        # Determine the masking layer
+        mask = get_mask_layer(self.model._MAPS.get("basins"), self.data, data)
 
-__all__ = ["StaticmapsComponent"]
+        # Set the data per layer
+        for dvar in data.data_vars:
+            if dvar in self._data:
+                logger.warning(f"Replacing grid map: {dvar}")
+            if mask is not None:
+                if data[dvar].dtype != "bool":
```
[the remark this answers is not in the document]
dalmijn: Bit of copied code from somewhere else in the package. Worked there so I didnt bother adjusting it. I think you're right though.

hydromt_wflow/components/config.py
```diff
@@ -133,8 +134,13 @@ def write(
             logger.info(f"Writing model config to {write_path.as_posix()}.")
             write_path.parent.mkdir(parents=True, exist_ok=True)
 
+            # Solve the pathing in the data
+            # Extra check for dir_input
+            rel_path = Path(write_path.parent, self.get_value("dir_input", fallback=""))
+            write_data = make_config_paths_relative(self.data, rel_path)
```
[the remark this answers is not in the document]
dalmijn: I doesnt work properly in core.

hydromt_wflow/wflow.py
```diff
@@ -719,18 +795,20 @@ def setup_floodplains(
             logger.debug(f"Using river_upa value value of: {new_river_upa}")
 
             # get river floodplain volume
-            inv_rename = {v: k for k, v in self._MAPS.items() if v in self.grid}
+            inv_rename = {
+                v: k for k, v in self._MAPS.items() if v in self.staticmaps.data
+            }
             da_fldpln = workflows.river_floodplain_volume(
                 ds=ds_hydro,
-                ds_model=self.grid.rename(inv_rename),
```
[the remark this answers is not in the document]
dalmijn: I talked it over and by just changing the calls, it should be fine.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,239 @@
...
+        ----------
+        filename : str, optional
+            Filename relative to model root, by default None
+        **kwargs : dict
+            Additional keyword arguments to be passed to the `read_nc` method.
+        """
+        # Supercharge with parent method
+        super().read(
+            filename=filename,
+            mask_and_scale=False,
```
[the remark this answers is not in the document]
dalmijn: Indeed.

hydromt_wflow/components/staticmaps.py
```diff
@@ -1,11 +1,239 @@
...
+            HydroMT model instance
+        filename : str
+            The path to use for reading and writing of component data by default.
+            By default "staticmaps.nc".
+        region_component : str, optional
+            The name of the region component to use as reference for this component's
+            region. If None, the region will be set to the grid extent.
+            Note that the create method only works if the region_component is None.
+            For add_data_from_* methods, the other region_component should be
+            a reference to another grid component for correct reprojection,
```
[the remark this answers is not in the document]
dalmijn: It does work, _get_grid_data will then just return itself.

hydromt_wflow/components/utils.py
dalmijn: Depends of the control flow I'd say, apart from that I fully agree.

deltamarnix (approved):
I think we can continue from here. Let's see if we can get those tests running soon.

PR #461, opened 2025-06-16
Feat: benchmarking

### LINKED ISSUE #441: Define and Implement Scenario Benchmarking for Model Build Time
To ensure consistent performance across releases, we need to define a benchmarking framework for model build time. This includes:

- [ ] Selecting a representative scenario (e.g., Moselle to be consistent with wflow representative scenario).
- [ ] Measuring the current model build time under standard conditions.
- [ ] Establishing an acceptable time range for future comparisons.
- [ ] Automating this benchmark as part of the CI/CD pipeline or performance testing suite.

savente93:
## Issue addressed
Fixes #441 

## Explanation

Couple of things that are noteworthy: 
1. I decided to use (pytest-benchmark)[https://pytest-benchmark.readthedocs.io/en/latest/index.html] since that integrates well with the current setup
2. I added some flags to the pyproject.toml to make sure the benchmarks don't get run when you do a test run (should work locally and in ci) as long as you use the pixi tasks for that. If you want to run the test bench outside of pixi just make sure to add `--benchmark-skip` to your command
3. Given that we'll only do this for releases (as confirmed by Ali) I think it is fine to keep the timing data in the repo itself. This shouldn't be anything sensitive if it is run from CI (it does include some information about the machine you run it on but only for comparison's sake) 
4. Because we wanna move on asap with this, I didn't do any fancy CI work. It is just manual workflows for now which I think is easiest, I can do more fancy stuff if necessary
5. For consistencies sake I thought it would be best to run the benchmark in CI (otherwise we'd need to ensure we have consistent specs for the person doing the release). This does make the actual committing of the snapshot a little more annoying so eventually I settled with having the CI job open a pull request to main with the timing data in it. Then you can change whatever branch you want it to go into later manually. The version used is determined against which branch you run the workflow. 
6. I haven't tested this very extensively, especially not in conjunction with the release flow, but locally it works all and I made sure it doesn't impact day-to-day developments I think this is a good compromise.

## Checklist


.github/workflows/compare-benchmark.yml
```diff
@@ -0,0 +1,32 @@
...
+  workflow_dispatch:
+
+
+jobs:
+  test:
+    defaults:
+      run:
+        shell: bash -l {0}
+    env:
+      branch_name: ${{ github.head_ref || github.ref_name }}
```
deltamarnix, line 13: I think that this env var is never used. Also in the other scripts.

.github/workflows/compare-benchmark.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Setup pixi env
+        uses: prefix-dev/setup-pixi@v0.8.8
+        with:
+          pixi-version: "v0.45.0"
```
deltamarnix, line 25: The other scripts are on an older version of pixi. Better to keep them all the same for now?

.github/workflows/compare-benchmark.yml
```diff
@@ -0,0 +1,32 @@
...
+          environments: test-full-py${{ matrix.python-version }}
```
deltamarnix, line 26: I might be missing something, but there is no matrix.python-version

.github/workflows/record-benchmark.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Setup pixi env
+        uses: prefix-dev/setup-pixi@v0.8.8
+        with:
+          pixi-version: "v0.45.0"
+          environments: test-full-py${{ matrix.python-version }}
```
deltamarnix, line 26: There is no matrix specified for python-version. So different python versions won't be picked up

pyproject.toml
deltamarnix, line 1: How come that so much changed on this file? It means that multiple people in the team have a different formatter for toml files.

.github/workflows/compare-benchmark.yml
savente93: You're right

.github/workflows/record-benchmark.yml
```diff
@@ -0,0 +1,42 @@
...
+          locked: false
+          cache: true
+
+      - name: Tests
+        run: |
+            pixi run -e test-full-py${{ matrix.python-version }} benchmark-record
+
+            git pull # make sure git doesn't reject our push because of unknown refs
```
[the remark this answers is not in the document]
savente93: well TIL

pyproject.toml
savente93: it's probably just me, I'll disable it and revert the unnecessary changes.

pyproject.toml
```diff
@@ -205,6 +212,57 @@ serve = { cmd = [
...
+clean-bin = { cmd = ["rm", "-rf", "bin"] }
+clean-dist = { cmd = ["rm", "-rf", "dist"] }
+clean-docs-build = { cmd = ["rm", "-rf", "docs/_build"] }
+clean-docs-example = { cmd = ["rm", "-rf", "docs/_examples"] }
+clean-docs-gen = { cmd = ["rm", "-rf", "docs/_generated"] }
+clean-docs = { depends-on = [
+  "clean-docs-build",
+  "clean-docs-example",
+  "clean-docs-gen",
+] }
```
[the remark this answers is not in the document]
savente93: I think that's fair, I think it got included by me resolving a merge conflict incorrectly. I'll remove it

deltamarnix (approved):

PR #469, opened 2025-06-23
Feat/add states component

### LINKED ISSUE #399: Create StatesComponent for HydroMT core v1.0
Create a StatesComponent that can read and write a netcdf file.

hboisgon:
## Issue addressed
Fixes #399 

## Explanation
Implemented `WflowStatesComponent` as a subclass of core `GridComponent`. Most methods are only a direct call to core, only the docstrings have been added.
Added in `WflowModel`, `set_states` as it used to be taken directly from core pre-v1.

I also added an example of states file to test reading in *wflow_piave_subbasin*.

The docs are still TODO but waiting for the config PR to be merged to get the right structure.

## Checklist

### DIFF
--- a/hydromt_wflow/utils.py
+++ b/hydromt_wflow/utils.py
@@ -475,6 +475,16 @@ def get_grid_from_config(
     return da
 
 
+def _mask_data_array(data_array: xr.DataArray, mask: xr.DataArray) -> xr.DataArray:
+    """Mask the data array based on the mask."""
+    # If the data is boolean, we set it to False where the mask is False
+    if data_array.dtype == "bool":
+        return data_array.where(mask, False)
+    # Otherwise we set it to nodata where the mask is False
+    else:
+        return data_array.where(mask, data_array.raster.nodata)
+
+
 def mask_raster_from_layer(
     data: xr.Dataset | xr.DataArray, mask: xr.DataArray
 ) -> xr.Dataset | xr.DataArray:
@@ -493,25 +503,16 @@ def mask_raster_from_layer(
     ----------
         data (xr.Dataset, xr.DataArray):
             The grid data containing the data that should be masked
-        layer_name (string):
-            mask that the data will be masked to. Values can be boolean or
-            numeric. Places where this layer is different than the rater nodata will be
-            masked in the other data variables
+        mask (xr.DataArray):
+            mask that the data will be masked to. Values can be boolean or numeric.
+            Places where this layer is different than the raster nodata will be
+            masked in the other data.
 
     Returns
     -------
         xr.Dataset, xr.DataArray: The grid with all of the data variables masked.
     """
     mask = mask != mask.raster.nodata
-
-    def _mask_data_array(data_array: xr.DataArray) -> xr.DataArray:
-        """Mask the data array based on the mask."""
-        # If the data is boolean, we set it to False where the mask is False
-        if data_array.dtype == "bool":
-            return data_array.where(mask, False)
-        # Otherwise we set it to nodata where the mask is False
-        else:
-            return data_array.where(mask, data_array.raster.nodata)
 
     if isinstance(data, xr.DataArray):
         data = _mask_data_array(data)
--- a/tests/components/test_states_component.py
+++ b/tests/components/test_states_component.py
@@ -22,7 +22,7 @@ def test_wflow_states_component_init(mock_model: MagicMock):
     assert len(component.data) == 0
 
 
-def test_wflow_states_component_ini_with_region(mock_model_staticmaps: MagicMock):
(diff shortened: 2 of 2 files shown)


hydromt_wflow/components/states.py
```diff
@@ -0,0 +1,135 @@
...
+        if not isinstance(data, (xr.Dataset, xr.DataArray)):
+            raise TypeError("Data must be an xarray Dataset or DataArray")
+
+        if self._region_component is not None:
+            # Ensure the data is aligned with the region component (staticmaps)
+            region_grid = self._get_grid_data()
+            if not data.raster.identical_grid(region_grid):
+                raise ValueError(
+                    f"Data grid must be identical to {self._region_component} component"
+                )
```
dalmijn, line 85: I was thinking a little bit about this, I indeed think an error is best as you can almost safely assume that the states are otherwise from a different model.  Maybe only when you have states from an older model that describes the same region, you might want to use it, but even then.. Don't think we should allow a reproject here. Just spilling out my thoughts.

hydromt_wflow/utils.py
```diff
@@ -503,21 +503,21 @@ def mask_raster_from_layer(
...
+    if isinstance(data, xr.DataArray):
+        data = _mask_data_array(data)
     else:
         for var in data.data_vars:
-            # nodata is required for all but boolean fields
-            if data[var].dtype != "bool":
-                data[var] = data[var].where(mask, data[var].raster.nodata)
-            else:
-                data[var] = data[var].where(mask, False)
+            data[var] = _mask_data_array(data[var])
```
dalmijn, line 520: Limit inner functions as mush a possible when it's not used as a decorator.
I'd suggest moving it to the outside.

hydromt_wflow/wflow.py
```diff
@@ -5721,69 +5728,75 @@ def write_forcing(
...
+        Parameters
+        ----------
+        data: xarray.DataArray or xarray.Dataset
+            new map layer to add to states
+        name: str, optional
+            Name of new map layer, this is used to overwrite the name of a DataArray and
+            ignored if data is a Dataset
+        """
+        if self._MAPS["basins"] in self.grid:
+            data = utils.mask_raster_from_layer(data, self.grid[self._MAPS["basins"]])
```
dalmijn, line 5797: We will still incorporate the masking function in core GridComponent right? Just so we know we can move this later on.

tests/components/test_states_component.py
```diff
@@ -0,0 +1,175 @@
...
+    # Assert that the internal data is None
+    assert component._data is None
+
+    # When asking for data property, it should return a xarray.Dataset
+    assert isinstance(component.data, xr.Dataset)
+    assert isinstance(component._data, xr.Dataset)  # Same for internal
+    assert len(component.data) == 0
+
+
+def test_wflow_states_component_ini_with_region(mock_model_staticmaps: MagicMock):
```
dalmijn, line 25: Again, nitpicking.

```suggestion
def test_wflow_states_component_init_with_region(mock_model_staticmaps: MagicMock):
```

tests/components/test_states_component.py
```diff
@@ -0,0 +1,175 @@
...
+        component.set(array)
+
+    # Try with a different grid that is not identical
+    # This should raise an error
+    different_grid = grid_dummy_data.copy()
+    different_grid["x"] = different_grid.x + 1  # Shift x coordinates
+    with pytest.raises(
+        ValueError, match="Data grid must be identical to staticmaps component"
+    ):
+        component.set(different_grid)
```
dalmijn, line 84: Bit nitpicky, but checking for errors make for a great separate test.

tests/components/test_states_component.py
```diff
@@ -0,0 +1,175 @@
...
+    assert "river_instantaneous_q" in component.data
+
+
+def test_wflow_states_component_write(
+    mock_model: MagicMock,
+    grid_dummy_data: xr.DataArray,
+):
+    # Setup the component
+    component = WflowStatesComponent(mock_model)
+
```
dalmijn, line 135: Very nitpicky, delete this empty line.

dalmijn (changes requested):
@hboisgon, great work! Testing is really well done. That's nice to see :).
Only have some really nitpicky comments.

dalmijn (approved):
Discussed, changes will be made, approved therefore in advance. LGTM.

hydromt_wflow/wflow.py
hboisgon: I think so yes. Although there is a similar mask function in core but that does not deal with boolean like we do.: https://github.com/Deltares/hydromt/blob/35586496c8c97d4f46c2b6429d4756544f3cf376/hydromt/gis/raster.py#L1922-L1932
Would be nice to merge functionnality indeed in core.

PR #490, opened 2025-07-03
fix: ensure glaciers are reprojected to model crs

### LINKED ISSUE #467: Incorrect glaciermaps when glacier data and model have a different CRS

### Reproducible Example

When the model has a different CRS than the glacier dataset, the geometry intersection in the glacier workflow does not work. Despite the geometries being at the same geographical location, the intersect algorithm is not aware of this. See this line: https://github.com/Deltares/hydromt_wflow/blob/382bb9aee7a77bff31027fc18dcb41a676a7ed41/hydromt_wflow/workflows/glaciers.py#L103

### Current behaviour

See above. As a result, the glacierfrac and glacierstore maps are not populated with values. They are added to the model, but only contain missing values.

### Desired behaviour

Probably best to reproject the glacier dataset to the model CRS, in order to properly perform the intersection.

For example, like this:
```python 
   ....

    # reproject to match ds_like CRS
    gdf = gdf.to_crs(ds_like.raster.crs)

    for i in range(len(gdf)):
        glacier = gdf.iloc[i]
        gridded_glacier = gdf_grid.intersection(glacier.geometry)
```

### Additional context

_No response_

LuukBlom:
## Issue addressed
Fixes #467 

## Explanation
- Before checking for intersections with the glaciers, convert to the correct crs.
- Added unit tests for the `glaciermaps` workflow
- Added integration test with 2 `.geojson` files with different crs

## Checklist

I created the geojson files in QGIS, ensuring they overlap with the `.tif` files located in the `merit_utm.yml` datacatalog, then exported to different crs.


LuukBlom:
> Nice work @LuukBlom! Just two comments on the tests and the data catalog

Thanks! I can only see one comment about the data catalog though?

LuukBlom:
> > > Nice work @LuukBlom! Just two comments on the tests and the data catalog
> > 
> > 
> > Thanks! I can only see one comment about the data catalog though?
> 
> I wrote a comment on the test_model_methods.py, but that apparently dissappeared. I suggested you also make sure the CRS is the expected CRS for the glaciermaps in the test_projected_crs_glaciers test

Ah yes, good idea!

PR #504, opened 2025-07-15
fix: re-enable all integration tests again

### LINKED ISSUE #347: Tests should be renabled for v1 branch when possible.
### Kind of request

### Enhancement Description

Currently the tests for v1 will be broken because of partial v1 addoption, but we should strive to renable and fix them asap 

### Use case

_No response_

### Additional Context

_No response_

LuukBlom:
## Issue addressed
Fixes #347 

## Explanation
self-explanatory

## Checklist


LuukBlom:
All tests are green!

Except for `test_build_model`, which fails in the `_compare_wflow_models` check with only 1 invalid map!
I uncommented that line and the rest of the comparison doesnt raise any more asserts, so this is truly the last thing to resolve. 

This is the error: 

`meta_subgrid_outlet_idx`: 'mean diff (1421 cells): -717706.0542; '

I plotted both maps and the problem is clear.
I applied the mask in the test using the code below (and in [this temporary commit](https://github.com/Deltares/hydromt_wflow/pull/504/commits/88c70025cbf765ce7dae35801d4dd56979c2389f)) when the assertion failed, the results look correct:

(left is example model, middle is model built in the test, right is middle one masked with the first one)
image

```python
def  _compare_wflow_models(mod0, mod1):
    ...
    vmin = min(map0.min().item(), map1.min().item())
    vmax = max(map0.max().item(), map1.max().item())

    plt.figure()
    map0.plot(vmin=vmin, vmax=vmax, cmap="viridis")

    plt.figure()
    map1.plot(vmin=vmin, vmax=vmax, cmap="viridis")

    plt.figure()
    masked = mod0.staticmaps.data[name].where(mod1.staticmaps.data[name])
    masked.plot(vmin=vmin, vmax=vmax, cmap="viridis")

    plt.show()
    ...
```

So somewhere there is a mask not being applied, but Im struggling find out where. The only place where `meta_subgrid_outlet_idx` is mentioned is in setup_basemaps, but when debugging there, all maps seem correct.

hboisgon:
The change in `test_setup_pet_forcing` is likely due to this issue: https://github.com/Deltares/hydromt/issues/1226

I also really do not like that the buffer becomes so dependent on the original data resolution and would like to go back to a buffer in pixel for rasters and not meters

hboisgon:
When fixing the reading as mentioned above, it turns out clipping is not working anymore (forcing and states are finally read this time). Part of it is because if you are using all files forcing/states could have different names for the dims and non-flipped latitudes (we used to only write with gdal_complaint, rename_dims and force_sn for staticmaps and not the other components).
But for clipping the issue is that when clipping forcing the unclipped forcing cannot be read anymore because its grid will be different from the already clipped grid in staticmaps. 
I guess the solution depends on how do we fix the clip command / function in hydromt_wflow. We may have to do one clip function that does all but if you have several forcing file, it means you will always need to clip the staticmaps first once again.

LuukBlom:
I ran the skipped tests with an editable install of hydromt with the `buffer` changes: https://github.com/Deltares/hydromt/pull/1245

And all tests are green with the old buffer behaviour! (`test_model_build`, and the setup_method tests)

For now I will merge this PR so we stop scope creeping this one.

PR #513, opened 2025-08-01
Feature: Forcing Component

### LINKED ISSUE #398: Create WflowForcingComponent for HydroMT core v1.0
Create a class ForcingComponent that can read and write a forcing netcdf file.

dalmijn:
## Issue addressed
Fixes #398 

## Explanation
Created `WflowForcingComponent` in which I tried to port over the `write` method as close to the original as possible while also fixing and adding features to HydroMT-core. The  `WflowForcingComponent` is linked spatially to the `WflowStaticmapsComponent`. However, in order to prevent the new component from erroring all over the place, it reprojects the forcing data on the fly when the forcing grid does not spatially match the staticmaps grid. This however can only be done when there is a data present in the `WflowStaticmapsComponent`. When the `WflowStaticmapsComponent` has no data, the `WflowForcingComponent` is still allowed to read and set data, but will not check. Checking resumes when data starts to appear in the  `WflowStaticmapsComponent` and if not equal, it will reproject when `data` is called.

edit:
reprojecting is no longer done in `set()` or the `data` property. It is the responsibility of the caller (workflow functions most of the time) to ensure the grids (staticmaps & forcing) are identical, `set()` will error when they are not, and data just initializes & returns data added to the model using `set()`.

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -133,7 +133,7 @@ Other general methods
    WflowModel.set_config
    WflowModel.get_config
    WflowModel.set_forcing
-   WflowModel.set_staticmaps
+   WflowModel.set_grid
    WflowModel.set_tables
    WflowModel.set_states
 
@@ -241,7 +241,7 @@ Other general methods
    WflowSedimentModel.set_config
    WflowSedimentModel.get_config
    WflowSedimentModel.set_forcing
-   WflowSedimentModel.set_staticmaps
+   WflowSedimentModel.set_grid
    WflowSedimentModel.set_states
 
 .. _data_containers:
--- a/tests/test_model_methods.py
+++ b/tests/test_model_methods.py
@@ -541,7 +541,7 @@ def test_setup_outlets(example_wflow_model):
     # Update subcatchment ID
     new_subcatch = example_wflow_model.grid["subcatchment"].copy()
     new_subcatch = new_subcatch.where(new_subcatch == new_subcatch.raster.nodata, 1001)
-    example_wflow_model.set_staticmaps(new_subcatch, "subcatchment")
+    example_wflow_model.set_grid(new_subcatch, "subcatchment")
 
     # Derive outlets
     example_wflow_model.setup_outlets()


hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,280 @@
...
+            if reference_component.data is None:
+                raise ValueError(
+                    f"Unable to get grid from the referenced region component: \
+'{self._region_component}'."
+                )
+            return reference_component.data
+
+        if self._data is None:
+            raise ValueError("Unable to get grid data from this component.")
+        return self._data
```
LuukBlom, line 74: @hboisgon, this is the change where we need to get out of the infinite loop:

self.data -> self._reproj_data() -> self.get_grid_data() -> self.data -> ...

LuukBlom: I moved the checks from the property `self.data` to `set()`

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,252 @@
...
+        **kwargs : dict
+            Additional keyword arguments to be passed to the `write_nc` method.
+        """
+        super().read(
+            filename=filename,
+            mask_and_scale=False,
+            **kwargs,
+        )
+
+    def write(
```
hboisgon, line 105: There is a whole bunch of code that disappeared that was dealing with what if the output filename already exist. It can take hours to produce a wflow forcing file (unfortunately) and it would be a shame to loose hours because users forgot to update the output forcing filename. What we used to do:

1. If filename does not already exist just write
2. Else prepare a new filename (based on attributes like data source and start/end year)
3. If names from step 1 and step 2 already exist, just throw a warning and skip writting.

I still think it is useful to prevent overwritting existing forcing file. Maybe we can skip step 2 or use a less complex name (ie less code) or add an allow overwrite flag in the argument which would be False by default?

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,252 @@
...
+
+        return filename, starttime, endtime
+
+    ## Mutating methods
+    def set(
+        self,
+        data: xr.DataArray | xr.Dataset,
+        name: str | None = None,
+    ):
+        """Add data to grid.
```
hboisgon, line 229:
```suggestion
        """Add data to forcing.
```

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,252 @@
...
+
+        All layers of grid must have identical spatial coordinates.
```
hboisgon, line 231:
```suggestion
        All layers in data must have identical spatial coordinates to existing forcing/staticmaps.
```

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,252 @@
...
+        name: str, optional
+            Name of new map layer, this is used to overwrite the name of a DataArray
+            and ignored if data is a Dataset
+        """
+        # Rename the temporal dimension to time
+        if "time" not in data.dims:
+            raise ValueError("'time' dimension not found in data")
+
+        # Check spatial extent
+        data = self._reproj_data(data)
```
hboisgon, line 246: This should be more than enough:

```python
        if self._region_component is not None:
            # Ensure the data is aligned with the region component (staticmaps)
            region_grid = self._get_grid_data()
            if not data.raster.identical_grid(region_grid):
                raise ValueError(
                    f"Data grid must be identical to {self._region_component} component"
                )
```

hboisgon (changes requested):

hydromt_wflow/components/forcing.py
LuukBlom: I added the old code back, but it is indeed quite some code for generating a new filename.

Alternatively, we could also just append `datetime.now()` to the filename to always generate a unique name? 

This would always write the file and include some metadata on when it was created for users to differentiate between the files.

pyproject.toml
```diff
@@ -41,11 +41,11 @@ classifiers = [
 ]
 
 # Core dependencies
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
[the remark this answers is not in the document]
LuukBlom: In hydromt core [this PR](https://github.com/Deltares/hydromt/pull/1215) Brendan also increased the minimum python version, where he cited [this site](https://scientific-python.org/specs/spec-0000/).
This sounds good to me tbh!

hydromt_wflow/wflow.py
```diff
@@ -5514,191 +5503,52 @@ def write_forcing(
...
+            starttime=self.config.get_value("time.starttime"),
+            endtime=self.config.get_value("time.endtime"),
+            time_chunk=time_chunk,
+            time_units=time_units,
+            **kwargs,
+        )
+        # Set back to the config
+        self.config.set("input.path_forcing", filename.as_posix())
+        self.config.set("time.starttime", starttime)
+        self.config.set("time.endtime", endtime)
```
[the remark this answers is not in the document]
LuukBlom: I added `_validate_timespan()` to the ForcingComponent, which gets called before writing.

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,252 @@
...
+class WflowForcingComponent(GridComponent):
+    """Wflow forcing component."""
+
+    def __init__(
+        self,
+        model: Model,
+        *,
+        filename: str = "inmaps.nc",
+        region_component: str | None = None,
+        region_filename: str = "geoms/forcing_region.geojson",
```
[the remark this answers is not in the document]
LuukBlom: I think setting these to None will raise an error at some point.
The value of `region_filename` gets passed down into [`SpatialModelComponent`](https://github.com/Deltares/hydromt/blob/e9416417205fcca1cb7d73da816c0d2a8b1fab50/hydromt/model/components/spatial.py#L22), where it is stored in `self._region_filename`. 
This is only used in the fn `write_region()` where it is used without any checks:

```python
    def write_region(
        self,
        *,
        filename: Optional[StrPath] = None,
        to_wgs84=False,
        **write_kwargs,
    ) -> None:
        self.root._assert_write_mode()
        if self._region_component is not None:
            logger.info(
                "Region is a reference to another component. Skipping writing..."
            )
            return

        if self.region is None:
            logger.info("No region data available to write.")
            return

        _write_region(
            self.region,
            filename=filename or self._region_filename,
            to_wgs84=to_wgs84,
            root_path=self.root.path,
            **write_kwargs,
        )
```

So if `filename=None`, this fn will error. 
I would prefer to just write a random region file that the user can then delete.

pyproject.toml
```diff
@@ -41,11 +41,11 @@ classifiers = [
...
 dependencies = [
   "dask",
   "geopandas>=0.10",
-  "hydromt>=1.0.1",
+  "hydromt@https://github.com/Deltares/hydromt.git", # TODO Set back after hydromt 1.2.1
```
deltamarnix, line 48: What did you need from a newer hydromt core? Do you want to release hydromt core before merging this PR?

hydromt_wflow/components/forcing.py
hboisgon: In that case I would leave it up to the user to say that overwritting the existing file is allowed. I'm not sure how much we can guess unless the file is completely empty or corrupted. 
I would by default in the code skip, but in our build template yml switch to allow for overwritting

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,358 @@
...
+        """
+        Initialize a WflowForcingComponent.
+
+        Parameters
+        ----------
+        model : Model
+            HydroMT model instance
+        filename : str, optional
+            Default path relative to the root where the forcing file(s) will be read
+            and written. By default 'inmaps.nc'. By default "inmaps.nc".
```
hboisgon, line 46:
```suggestion
            Default path relative to the root where the forcing file(s) will be read
            and written. By default 'inmaps.nc'.
```

hydromt_wflow/wflow.py
```diff
@@ -2965,7 +2963,7 @@ def setup_lulcmaps_with_paddy(
             logger=logger,
         )
         rmdict = {k: self._MAPS.get(k, k) for k in landuse_maps.data_vars}
-        self.set_grid(landuse_maps.rename(rmdict))
+        self.set_staticmaps(landuse_maps.rename(rmdict))
```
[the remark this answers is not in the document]
LuukBlom: I agree, will revert these changes

hydromt_wflow/components/forcing.py
```diff
@@ -0,0 +1,252 @@
...
+        ----------
+        model : Model
+            HydroMT model instance
+        filename : str, optional
+            The path to use for reading and writing of component data by default.
+            By default "inmaps.nc".
+        region_component : str, optional
+            The name of the region component to use as reference for this component's
+            region. If None, the region will be set to the grid extent. Note that the
+            create method only works if the region_component is None.
```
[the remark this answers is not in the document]
LuukBlom: I see why it was added: GridComponent in core had a `create_from_region` method before [this PR](https://github.com/Deltares/hydromt/pull/1229) was merged. 
Which did this check: 
```python
if self._region_component is not None:
    raise ValueError(
        "Region is a reference to another component. Cannot create grid."
    )
``` 

So no longer relevant and I will remove

hydromt_wflow/components/forcing.py
hboisgon: No it's something different. Again for forcing we want to force file protection compared to a normal model update where overwritting staticmaps is allowed.

hydromt_wflow/wflow.py
```diff
@@ -5458,47 +5459,35 @@ def read_forcing(self):
...
         """
-        fn_default = "inmaps.nc"
-        fn = self.get_config(
-            "input.path_forcing", abs_path=True, fallback=join(self.root, fn_default)
+        # Sort which path/ filename is actually the one used
+        # Hierarchy is: 1: signature, 2: config, 3: default
+        p = (
+            filename
+            or self.config.get_value("input.path_forcing")
+            or self.forcing._filename
```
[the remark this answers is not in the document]
LuukBlom: Just had a discussion with Sam about this and this seems to be the least bad way to do this at the moment. 

forcing.read() takes a filename relative to the model root. (direct usage from hydromt core.)

In hydromt wflow however, we need to somehow include `dir_input` into the relative path (`p_input`).
If the fn argument `filename` and the config `input.path_forcing` are both None, but `dir_input` is defined, we need to construct the relative path using the private attr (which would be `model.root` / `dir_input` / `forcing._filename`). If we wouldnt do that and just pass None as the filename, we lose `dir_input`.

Might be an idea to make the `_filename` attr public?

PR #515, opened 2025-08-04
Refactor of reservoirs and lakes in Wflow.jl

### LINKED ISSUE #435: updating water bodies with setup_lakes always overwrites the model grid
When updating a wflow model with additional water bodies, the `setup_lakes` method overwrites the existing "lakes", and as a result only the new ones remain in the model. Include an option to append the new ones, or simply always append them, similar to `setup_gauges`.

### LINKED ISSUE #459: Change lake and reservoir setup (1.0 rc)
For the Wflow kernel (1.0 rc) we are restructuring the lake and reservoir structs (see also https://github.com/Deltares/Wflow.jl/issues/353 and https://github.com/Deltares/Wflow.jl/pull/634). Together with @JoostBuitink  and @hboisgon we have decided to use the term `reservoir` for both reservoirs and lakes.

Below an example of the TOML settings: 

```toml
[state.variables]
reservoir_water_surface__instantaneous_elevation = "waterlevel_reservoir"

[input]
reservoir_area__count = "wflow_reservoirareas"
reservoir_location__count = "wflow_reservoirlocs"

[input.static]
reservoir_surface__area = "reservoir_area"
"reservoir_water_demand~required~downstream__volume_flow_rate" = "ResDemand"
reservoir_water_release-below-spillway__max_volume_flow_rate = "ResMaxRelease"
reservoir_water__max_volume = "ResMaxVolume"
"reservoir_water~full-target__volume_fraction" = "ResTargetFullFrac"
"reservoir_water~min-target__volume_fraction" = "ResTargetMinFrac"
reservoir_water_surface__initial_elevation = "waterlevel_reservoir"
reservoir_water__rating_curve_type_count = "outflowfunc"
reservoir_water__storage_curve_type_count = "storfunc"
```

For the simple reservoir functionality also these variables need to be set: 
- `reservoir_water__rating_curve_type_count` (4)
- `reservoir_water__storage_curve_type_count` (1)
- `reservoir_water_surface__initial_elevation` (fullfrac * max_volume / area)

hboisgon:
## Issue addressed
Fixes #459 
Fixes #435 

## Explanation
Reservoirs and lakes have been merged into one structure in Wflow.jl rc2. The following have been changed:

### Code structure

- [x] Keep two separated functions for uncontrolled reservoirs (lakes, weirs) and simple control reservoirs (current reservoirs)
- [x] Rename ``setup_lakes`` to ``setup_reservoirs_no_control`` for sbm (hesitation with ``setup_reservoirs`` but may be confusing in the short term as this was the name of the other method) and to ``setup_natural_reservoirs`` for sediment. Includes renaming of the corresponding staticgeom. ``workflows.waterbodies.lakeattrs`` has been renamed to ``reservoir_parameters``.
- [x] Rename ``setup_reservoirs`` to ``setup_reservoirs_simple_control`` for sbm. Includes renaming of the corresponding staticgeom. ``workflows.waterbodies.reservoirattrs`` has been renamed to ``reservoir_simple_control_parameters``.
- [x] Create an extra staticgeom with all reservoirs (lakes + reservoirs) called reservoirs.geojson. The intermediate ones are still kept with their default name strating with meta_ (eg meta_reservoirs_simple_control) to facilitate updating a model for users.
- [x] Additional: restructure parts or full of ``WflowModel._setup_waterbodies`` into ``workflows.waterbodies.reservoir_id_maps`` (new name of ``workflows.waterbodies.waterbodymaps``). 
- [x] Rename ``workflows.waterbodies`` into ``workflows.reservoirs``
- [x] Moved all functions about conversion from Wflow version v0.x to v1 out of ``utils`` to a new ``version_upgrade`` script. This script can be moved later when we restructure our files.

### Functional edits

- [x] Update Wflow variable names for lakes in wflow.py, waterbodies.py and naming.py
- [x] Add three new variables for reservoir for the rating curve, storage curve and initial depth
- [x] Allow to merge lakes created by ``setup_lakes`` and reservoirs created by ``setup_reservoirs``. A new function has been created ``overwrite_existing`` with False as default behavior. The corresponding ``merge_reservoirs`` function has been added in ``workflows.waterbodies``. A check if done if IDs are not duplicated between lakes and reservoirs. For now returns either an error or a warning and skip updating (default) but we could instead change the IDs. Changing IDs means that we may loose ID mapping for hydroengine and gww and ID for rating curves csv tables names which is too complex for a case that we see may rarely happen.
- [x] Adapt the conversion of v0.x to v1.0: update the TOML and also update the staticmaps to merge both lakes and reservoirs and create additional parameters for simple control reservoir.
- [x] Update similarly for wflow_sediment
- [x] Update ``setup_cold_states``
- [x] Update ``clip_grid`` and ``clip_states``
- [x] Add a ``WflowModel.drop_vars_grid`` to remove variables from ``WflowModel.grid``

### Docs and testing

- [x] Update notebooks/examples
- [x] Update examples model (wflow_sbm, wflow_sbm clip, wflow_sediment)
- [x] Test if the new models run with Wflow: wflow_piave_subbasin and wflow_sediment_piave_subbasin

## Checklist


hydromt_wflow/workflows/reservoirs.py
LuukBlom, line 1: Secondly, as I've mentioned before, I would love a more structured/nested approach for the naming of the variables.
(Not really for this PR but wanted to mention again)

Its quite error prone/intimidating/overwhelming to have these variables defined as strings everywhere.
I would much prefer something like this:

```python
class ReservoirParameter(Enum):
    """Variables describing the reservoirs in Hydromt Wflow"""

    area = "reservoir_area"
    """We can also add docstrings to explain variables like this"""

    initial_depth = "reservoir_initial_depth"
    rating_curve = "reservoir_rating_curve"
    storage_curve = "reservoir_storage_curve"
    max_volume = "reservoir_max_volume"
    target_min_fraction = "reservoir_target_min_fraction"
    target_full_fraction = "reservoir_target_full_fraction"
    demand = "reservoir_demand"
    max_release = "reservoir_max_release"
    b = "reservoir_b"
    e = "reservoir_e"
    outflow_threshold = "reservoir_outflow_threshold"
    lower_id = "reservoir_lower_id"
    trapping_efficiency = "reservoir_trapping_efficiency"
    area_id = "reservoir_area_id"
    outlet_id = "reservoir_outlet_id"

    @staticmethod
    def layers():
        return (
            ReservoirParameter.common()
            + ReservoirParameter.control()
            + ReservoirParameter.uncontrolled()
            + [ReservoirParameter.area_id, ReservoirParameter.outlet_id]
        )

    @staticmethod
    def common():
        return [
            ReservoirParameter.area,
            ReservoirParameter.initial_depth,
            ReservoirParameter.rating_curve,
            ReservoirParameter.storage_curve,
        ]

    @staticmethod
    def control():
        return [
            ReservoirParameter.max_volume,
            ReservoirParameter.target_min_fraction,
            ReservoirParameter.target_full_fraction,
            ReservoirParameter.demand,
            ReservoirParameter.max_release,
        ]

    @staticmethod
    def uncontrolled():
        return [
            ReservoirParameter.b,
            ReservoirParameter.e,
            ReservoirParameter.outflow_threshold,
            ReservoirParameter.lower_id,
        ]

    @staticmethod
    def sediment():
        return [
            ReservoirParameter.trapping_efficiency,
            ReservoirParameter.area,
            ReservoirParameter.area_id,
            ReservoirParameter.outlet_id,
        ]

    _JULIA_NAMES: dict[str, str] = {...}

    def to_julia_name(self):
        """Convert parameter name to Julia name."""
        return self._JULIA_NAMES.get(self.value, self.value)
```

Or some other solution using pydantic BaseModels. 

I'd like to make an issue for this, but also will accept it might be unnecessary work without adding a lot of new value (just nicer/easier to use/more stable imo). 
What do you think?

LuukBlom: This also means we can replace all occurences of the stringd `"reservoir_<var>"` in the code with `ReservoirParameter.<var>`, which would then make renaming  / changing them much easier as well. 

Additionally, users could look at the documented ReservoirParameter class to see and understand all the options they have. (how would they find out now?)

LuukBlom (approved):
Very nice job! 

As you said, a lot of file changes are data changes so not a lot to comment on there.

Code-wise, I dont have a lot to say besides some preferences and nitpicks. So very nicely done!
I have a comment about raising vs warning + returning None.
And a more general comment about how we handle the variable names.

.vscode/settings.json
```diff
@@ -13,5 +13,8 @@
     "sonarlint.connectedMode.project": {
         "connectionId": "deltares",
         "projectKey": "Deltares_hydromt_wflow"
-    }
+    },
+    "python-envs.defaultEnvManager": "ms-python.python:conda",
+    "python-envs.defaultPackageManager": "ms-python.python:conda",
+    "python-envs.pythonProjects": []
```
LuukBlom, line 19: These settings belong in your user settings: 

in vs-code: `Ctrl + ,`

examples/wflow_extend_sediment.yml
```diff
@@ -13,12 +13,15 @@ setup_config: # options parsed to wflow toml file <section>.<option>
...
-# Else you can use a default of 1 via setup_constant_pars below
+# # If you do have reservoirs, you can re-run setup_lakes and setup_reservoirs to add it
 # setup_reservoirs:
 #   reservoirs_fn: hydro_reservoirs  # source for reservoirs shape and attributes
 #   min_area: 1.0           # minimum lake area to consider [km2]
 #   trapping_default: 1.0   # default trapping efficiency for reservoirs [0-1]
+#
+# setup_lakes:
+#   lakes_fn: hydro_lakes   # source for uncontrolled reservoirs (e.g. lakes) based on hydroLAKES: {hydro_lakes}
+#   min_area: 10.0          # minimum reservoir area to consider [km2]
```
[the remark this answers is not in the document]
LuukBlom: nevermind, I hadnt seen sediment.py yet when I wrote this comment. 
Still relevant though, see: https://github.com/Deltares/hydromt_wflow/pull/515/files#r2260271194

hydromt_wflow/wflow_sediment.py
```diff
@@ -204,85 +205,44 @@ def setup_rivers(
...
-        ds_lakes["lake_area"] = self.grid.raster.rasterize(
-            gdf_points, col_name="Area_avg", dtype="float32", nodata=-999
+        self.setup_reservoirs(
+            reservoirs_fn=lakes_fn,
+            overwrite_existing=overwrite_existing,
+            min_area=min_area,
+            trapping_default=0.0,  # lakes have no trapping efficiency
+            output_names=output_names,
+            geom_name=geom_name,
+            **kwargs,
```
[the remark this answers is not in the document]
hboisgon: Yes I'm not really sure about the new names so maybe will ask a couple of people when they are back from holidays for advice. 
Problem is that in hydrology you want to separate reservoirs with operations that control flow release (reservoirs with gates) versus reservoirs that cannot adjust flow (natural lakes or weirs). 
In sediment, I want to control reservoirs that have a man made dam (reservoirs with gates or weirs) versus no man-made structure (natural lakes). In sediment the function for ``setup_reservoirs`` is also now very general and for expert users, ``setup_lakes`` is now unnecessary and more of a shortcut or easy way in for beginner wflow_sediment users. So I could also rename ``setup_lakes`` to ``setup_natural_reservoirs`` or so but lakes is prettier in a way.

hydromt_wflow/workflows/reservoirs.py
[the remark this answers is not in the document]
hboisgon: Yes I'm not 100% fond of it either. The problem of Wflow is that is starts being used in globally applicable automated tools like CST or Flood-Adapt. In these tools, I would think you prefer to skip lakes in your models rather than get an error that you cannot do anything about apart from updating the global data. But I agree that it may get hidden for other users.

I think I left in the description above that what I could do instead of returning None is just to come up with new IDs (using max+1 or just 1-len(reservoirs) for example). Or else add a flag that by default you return an error but then global tools can switch it off to a warning and skip.

hboisgon: I like this :) feel free to open the issue, I think it would still be beneficial to implement but needs some refinements. I think it may actually be something to harmonise together with core potentially.

In hydromt, the hydromt name is very important in order to share data catalogs or workflows within plugins.
Eg if you wanted to get the DEM of a sfincs or wflow model you could just do:
```python
model = SfincsModel(...) or model = WflowModel(...)
dem = model.grid[self._MAPS["elevtn"]]
```
Well we now have flexible components so we got rid of part of this possibility anyhow but this doesn't change for the data catalog. So I wonder if the "base" names should not be hydromt based rather than wflow (staticmaps or julia) based. Ie to read DEM with DataCatalog you should have something still standard for both sfincs and wflow I think:
```
from hydromt.naming import TopographyParameters
elevation = self.data_catalog.get_rasterdataset(..., variables=TopographyParameters.elevation) # but wow this gets long instead of just "elevtn"
```
and not
```
elevation = self.data_catalog.get_rasterdataset(..., variables=BaseParameters.dem.hydromt_name) # for wflow
elevation = self.data_catalog.get_rasterdataset(..., variables=TopographyParameters.elevation.hydromt_name) #for sfincs
```

Anyhow good to have a refinement session on this and then decide if it is worth it. I will not implement yet in this PR for reservoirs but feel free to copy your nice example in the new issue!

.vscode/settings.json
hboisgon: To be honest, sharing the .vscode files like settings did not work great for me and my whole testing environment got messed up (still not fully working as I cant debug tests only run...) so I would prefer to just put back the .vscode files into gitignore. But well I will erase for now.

LuukBlom: Hmm, that sucks. project level vscode settings are meant for this, and we copied launch.json the vscode docs: https://code.visualstudio.com/docs/python/testing#_debug-tests

Some questions so I can investigate:
Does the test discovery complete sucessfully? 
Have you selected the correct interpreter? (`Ctrl + Shift + P` -> `select interpreter`)
Can you share your user settings / launch configs?

hydromt_wflow/workflows/reservoirs.py
LuukBlom: https://github.com/Deltares/hydromt_wflow/issues/517

.vscode/settings.json
deltamarnix: I think we are all using pixi for development. In a different PR we can update the developer docs on how to setup the environment. It seems to reference make_env.py still: https://deltares.github.io/hydromt_wflow/stable/dev_guide/dev_install.html

hydromt_wflow/wflow.py
```diff
@@ -2203,158 +2291,86 @@ def setup_reservoirs(
...
-            "reservoirs" for reservoirs.geojson.
+            "reservoirs_simple_control" for reservoirs_simple_control.geojson.
         kwargs: optional
             Keyword arguments passed to the method
             hydromt.DataCatalog.get_rasterdataset()
+        """  # noqa: E501
+        # retrieve data for basin
+        self.logger.info("Preparing reservoir with simple control maps.")
+        if "predicate" not in kwargs:
+            kwargs.update(predicate="contains")
```
deltamarnix, line 2302: https://docs.python.org/3/library/stdtypes.html#dict.setdefault

```suggestion
        kwargs.setdefault("predicate", "contains")
```

hydromt_wflow/wflow.py
```diff
@@ -2203,158 +2291,86 @@ def setup_reservoirs(
...
+            ds_res = workflows.reservoirs.merge_reservoirs(
+                ds_res,
+                self.grid.rename(inv_rename),
             )
-            output_name = self._MAPS.get(name, name)
-            self.set_grid(da_res.rename(output_name))
-            self._update_config_variable_name(output_name, data_type="static")
+            # Check if ds_res is None ie duplicate IDs
+            if ds_res is None:
+                return
```
deltamarnix, line 2347: I'm not familiar with when the logger is being called, and when not. Do we have rules for which classes/functions should add information to the logger? This seems like a good point to log I believe.

hydromt_wflow/wflow.py
```diff
@@ -5253,6 +5283,23 @@ def set_grid(
         # fall back on default set_grid behaviour
         GridModel.set_grid(self, data, name)
 
+    def drop_vars_grid(self, names: list[str], errors: str = "raise"):
```
deltamarnix, line 5286: Should this function be available for outsiders or is it supposed to be a private function?

hydromt_wflow/wflow.py
```diff
@@ -6103,80 +6150,53 @@ def clip_grid(self, region, buffer=0, align=None, crs=4326, inverse_clip=False):
...
+                # remove states
+                if (
+                    self.get_config(
+                        "state.variables.reservoir_water_surface__instantaneous_elevation"  # noqa: E501
+                    )
+                    is not None
+                ):
+                    del self.config["state"]["variables"][
+                        "reservoir_water_surface__instantaneous_elevation"
+                    ]
```
deltamarnix, line 6190: Would it be an option to delete it anyways, even if it doesn't exist?

hydromt_wflow/wflow_sediment.py
```diff
@@ -204,85 +205,44 @@ def setup_rivers(
 
     def setup_lakes(
```
deltamarnix, line 206: Do lakes still exist in the sediment model or are we expecting this to change into reservoirs as well? What is Wflow team's rationale to keep the concept of lake in sediment, but not in sbm?

hydromt_wflow/wflow_sediment.py
```diff
@@ -336,13 +298,32 @@ def setup_reservoirs(
             Keyword arguments passed to the method
             hydromt.DataCatalog.get_rasterdataset()
         """
-        # Derive lake are and outlet maps
-        gdf_res, ds_res = self._setup_waterbodies(
-            reservoirs_fn, "reservoir", min_area, **kwargs
+        # retrieve data for basin
+        self.logger.info("Preparing reservoir maps.")
+        if "predicate" not in kwargs:
+            kwargs.update(predicate="contains")
```
deltamarnix, line 304:
```suggestion
        kwargs.setdefault("predicate", "contains")
```

deltamarnix (commented):
I mainly have some considerations to go through. Overall it seems like this is going to work fine.

We're lucky that wflow was only released as a release candidate, otherwise the conversion functions should have been split up into multiple steps. Keep in mind that someone who already upgraded their 0.x model to 1.0 can no longer make use of the function to upgrade their 1.0-rc1 to 1.0-rc2.

hydromt_wflow/workflows/reservoirs.py
```diff
@@ -650,38 +820,148 @@ def lakeattrs(
...
+                ds[layer] != ds[layer].raster.nodata, ds_like[layer]
+            )
+        else:
+            # all parameters for sediment should be in both
+            logger.warning(
+                f"Reservoir layer {layer} is not present in either the new dataset or"
+                "the wflow model. Skipping adding new reservoirs. Consider overwritting"
+                "to solve this issue."
+            )
+            return None
```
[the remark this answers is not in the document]
hboisgon: But they are not, already the first if statement is different

hydromt_wflow/wflow.py
```diff
@@ -1935,109 +1939,137 @@ def setup_areamap(
...
+        # Derive reservoir area and outlet maps
+        ds_reservoirs, gdf_org = workflows.reservoir_id_maps(
+            gdf=gdf_org,
+            ds_like=self.grid,
+            min_area=min_area,
+            uparea_name=self._MAPS["uparea"],
+            logger=self.logger,
+        )
+        if ds_reservoirs is None:
+            # No reservoirs of sufficient size found
```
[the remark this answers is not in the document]
hboisgon: The workflow function already sends one so it would be duplicated. In the workflow I can send a more precise warning like which IDs are duplicated so why it's None, while at wflow level, the function would need to return the duplicate IDs. But I can split it

hydromt_wflow/wflow.py
hboisgon: We do not really have official rules no. The logger is called in the workflows with more info than I could get at the wflow model (ie the duplicated IDs values) and I don't know if duplicating logger is nice? For this one I could split the message so I did

hydromt_wflow/wflow.py
hboisgon: I guess in a way same level as set_grid? I dont really like users to have to modify self._grid directly in order to remove some variables from their grid. This is advanced user / plugin developer though, same as for set_grid

hydromt_wflow/wflow_sediment.py
hboisgon: Talking with Joost about this. There is a difference as mentioned also in a previous comment to Luuk. Sediment cares about is there a barrier or not (ie natural reservoir or not) while sbm cares about are the outflows controlled. Weirs have a man made barrier but no outlfow control so it would be in one category in sbm and another for sediment.

hydromt_wflow/workflows/reservoirs.py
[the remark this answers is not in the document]
hboisgon: Had a chat with Joost but changing the IDs might give issue with ID mapping to get timeseries from hydroengine and gww and if we have rating curve csv tables we also need to update the tables object. So this seems a little complex for something that may barely ever happen. If I'm wrong and the problem is common then we can implement. In the meantime I added a flag to switch between error or skip behavior.

hydromt_wflow/wflow.py
hboisgon: HUm actually it would raise an error if not in config so does not yet solve my issue :)

https://github.com/Deltares/hydromt_wflow/blob/2e78c7bed5c4fd6dc449af3838c764345212c76d/hydromt_wflow/wflow.py#L5892-L5893

PR #533, opened 2025-08-27
Add results component

### LINKED ISSUE #528: Add results component
### Kind of request

None

### Enhancement Description

Add results component to WflowModel. Keep dict of xarray ie hydromt core ``DatasetsComponent``

### Use case

_No response_

### Additional Context

_No response_

hboisgon:
## Issue addressed
Fixes #528

## Explanation
Results used to be one object (dict of xarray objects). WIth the flexible component structure, I decided to split results into one component per file:

- ``output_grid`` for the netcdf gridded output. It then inherits from the GridComponent with staticmaps as its region.
- ``output_scalar`` for the netcdf scalar output. I realise we do not have a simple xarrayDataset components, only the dict of Datasets in core so I started from the base ModelComponent.
- ``output_csv`` for the csv output. This one is then the DatasetsComponent of core and objects within are hydromt.vector.GeoDataArray. To be able to add georeference, the staticmaps need to be passed so I added a ``locations_component`` attribute. It's not really the region as csv can contain outputs for 3 points withint the basin so I thought defining another name was better.

## Checklist
- [x] Branch is up to date with `feat/migration-hydromt-core-v1`

In the current v0 version, csv output are converted to several netcdf files and geo-references are also added in the process. For netcdf_scalar, we only read the file directly but in a way it could also be split to several GeoDataArrays in the same way as the csv output. Maybe still something we would like to do?

### DIFF
--- a/hydromt_wflow/components/output_grid.py
+++ b/hydromt_wflow/components/output_grid.py
@@ -115,7 +115,7 @@ class WflowOutputGridComponent(GridComponent):
         if not isinstance(data, (xr.DataArray, xr.Dataset)):
             raise TypeError("Data must be an xarray Dataset or DataArray")
 
-        # Rename the temporal dimension to time
+        # Check that time dimension is present
         if "time" not in data.dims:
             raise ValueError("'time' dimension not found in data")
 
--- a/tests/components/test_output_scalar_component.py
+++ b/tests/components/test_output_scalar_component.py
@@ -57,11 +57,34 @@ def test_wflow_output_scalar_component_set(
     assert len(component.data) == 1
 
 
+def test_wflow_output_scalar_component_set_alt(
+    mock_model: WflowModel, forcing_layer: xr.DataArray
+):
+    # Setup the component
+    component = WflowOutputScalarComponent(mock_model)
+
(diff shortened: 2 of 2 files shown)


LuukBlom (approved):
Very nice work! Code and tests look good, no comments at all!

Possibly, we could consider adding a sort of aggregated results component that uses the 3 components created in this pr to be more similar to the v0 implementation for users.

But to be honest, im not a big fan, I just want to have some conversation on this PR ;)

hydromt_wflow/components/output_scalar.py
```diff
@@ -0,0 +1,168 @@
...
+        assert self._data is not None
+
+        if isinstance(data, xr.DataArray):
+            # NOTE _name can be different from _data.name !
+            if name is not None:
+                data.name = name
+            elif name is None and data.name is not None:
+                name = data.name
+            elif data.name is None and name is None:
+                raise ValueError("Name required for DataArray.")
```
Tjalling-dejong, line 121: This code is not covered by your tests. Its quite trivial but still nice to cover.

Tjalling-dejong (approved):
Good job on the new output components. The code is clear, neat, and well documented. I have only two minor comments.

PR #534, opened 2025-08-27
unskipped setup_gauges and fixed issue with csv file reading

Tjalling-dejong:
## Issue addressed
Fixes https://github.com/Deltares/hydromt/issues/1243

Part of process of unblocking #347 

## Explanation
data_catalog.get_geodataframe() was missing a driver argument when using a gauges table file (.csv, .xlsx, etc.). By default get_geodataframe uses the pyogrio driver to read the data while it should be a geodataframw_table driver.

I also updated the update_model_gauges.ipynb notebook to v1.

## Checklist


examples/update_model_gauges.ipynb
```diff
@@ -279,8 +270,8 @@
     "# initialize image with geoaxes\n",
     "fig = plt.figure(figsize=figsize)\n",
     "ax = fig.add_subplot(projection=proj)\n",
-    "bbox = mod.grid.raster.box.to_crs(3857).buffer(5e3)\n",
-    "extent = np.array(bbox.to_crs(mod.grid.raster.crs).total_bounds)[[0, 2, 1, 3]]\n",
+    "bbox = mod.staticmaps.data.raster.box.to_crs(3857).buffer(5e3)\n",
+    "extent = np.array(bbox.to_crs(mod.staticmaps.data.raster.crs).total_bounds)[[0, 2, 1, 3]]\n",
```
hboisgon, line 274: We have a nice shortcut :)
```suggestion
    "extent = np.array(bbox.to_crs(mod.crs).total_bounds)[[0, 2, 1, 3]]\n",
```

examples/update_model_gauges.ipynb
```diff
@@ -194,7 +192,7 @@
    "outputs": [],
    "source": [
     "# NOTE: copy this line (without !) to your shell for more direct feedback\n",
-    "! hydromt update wflow wflow_piave_gauges -c setup_gauges --opt gauges_fn=test_stations.csv --opt basename=test-rain --opt snap_to_river=False --opt gauge_toml_header=\"['P-station']\" --opt gauge_toml_param=\"['atmosphere_water__precipitation_volume_flux']\" -v"
+    "! hydromt update wflow wflow_piave_gauges -i wflow_update_gauges_without_snapping.yml -v"
```
hboisgon, line 195: Nice that you created the corresponding yaml file! Could you add a cell to print their contents though? So that the users can see what you are explaining. You can check other notebooks for examples.

hydromt_wflow/wflow.py
```diff
@@ -1806,11 +1806,21 @@ def setup_gauges(
...
+            # Add driver argument when file is a table
+            ext = Path(gauges_fn).suffix
+            if ext in [".csv", ".parquet", ".xls", ".xlsx"]:
+                driver_dict = {"driver": {"name": "geodataframe_table"}}
+                kwargs.update(driver_dict)
+
             gdf_gauges = self.data_catalog.get_geodataframe(
                 gauges_fn,
                 geom=self.basins,
-                # assert_gtype="Point", hydromt#1243
```
hboisgon, line 1813: This is still an open issue then no? Or do we decide the plugins have to do this check now?

hydromt_wflow/wflow.py
```diff
@@ -1806,11 +1806,21 @@ def setup_gauges(
...
+            if "metadata" in kwargs:
+                kwargs["metadata"].update(crs=code)
+            else:
+                kwargs.update(metadata={"crs": code})
+
+            # Add driver argument when file is a table
+            ext = Path(gauges_fn).suffix
+            if ext in [".csv", ".parquet", ".xls", ".xlsx"]:
+                driver_dict = {"driver": {"name": "geodataframe_table"}}
+                kwargs.update(driver_dict)
```
hboisgon, line 1819: Nice! I feel like actually this should be merged with the crs. Ie we only want to modify the CRS for if the file extension is csv etc. and not geojson, shp. And then you can remove the comment # hydromt#1243 above :)

hboisgon (commented):
Nice! Thanks for digging up @Tjalling-dejong 

I'm not sure I'm happy about the solution though... I feel like this is not something straighforward for the users to pass the driver argument. Also the syntax is not so fun and I'm pretty sure not well documented...
Anyhow I now realised we really changed something in v1. In v10, we had both a default driver and then default drivers linked to specific file extensions. See for example:

https://github.com/Deltares/hydromt/blob/e5c93aed382065c6db60dd0d0f6d100eaa0c46bd/hydromt/data_adapter/geodataframe.py#L41-L48

But this is true for all data source types (also rasterdataset etc). So I for example guess that if I pass only a link to a netcdf file, I'll have to specify manually the driver like what is done here. Personnally I would in favor to update v1 to do what it used to: try to infer driver from file extension (with good debug message that driver not found, assuming ...). What do you think?

Tjalling-dejong:
@hboisgon 

At first I did agree with you on the inferring of the driver by hydromt core, but after looking through the core code I would say it is the plugins responsibility to pass on the driver argument. The drivers in hydromt core are set up in such a way that you can add your own custom drivers. Core can then not know which custom driver belongs to which file type.  I therefore think that is the plugins responsibility to pass on which driver is to be used.

For this particular case with setup_gauges I can include an optional driver key word argument that if not specified the driver is inferred from the file extension. See latest commit

examples/.update_model_gauges.ipynb.swp
savente93, line 1: I'm guessing this was added by accident

savente93 (approved):
except the swap file LGTM, just remove it before you merge pls

PR #549, opened 2025-09-10
Fix/clip v1

### LINKED ISSUE #317: Clipping wflow does NOT clip initial states (and does not automatically add outlet gauge?)

### Reproducible Example

(I can share an example on the Nile basin)

### Current behaviour

When clipping a wflow model the instates are not clipped. Instead, they are copied and pasted as unclipped versions, which cannot be used by the model. Also, a gauge is not automatically placed at the outlet of the clipped model. Users must manually run an update command `-c setup_outlets -v` to address this.

### Desired behaviour

instates.nc file must be clipped as well. Consider also placing a gauge automatically at the outlet of the clipped (sub)basin

### Additional context

_No response_

hboisgon:
## Issue addressed
Fixes #317 

## Explanation
There is now one large WflowModel.clip function that clips staticmaps/geoms/states and updates the config. For WflowSbmModel, tables are also updated.

For staticmaps/forcing/states, component level clip functions were added. For geoms, too many model methods would need to be called (for basins, rivers and outlets) so I decided not to have a clip function unless we decide that it is okay to call WflowModel functions in the component.

Because I wanted to test the states, I updated our compare_model_functions and tranfered to the component.test_equal and component.utils (testing for staticmaps and states are the same and the funciton we had is better than the core GridComponent.test_equal, more informative).

## Checklist
- [ ] Updated tests or added new tests -> TODO maybe still add some unit test for staticmaps/forcing.states.clip although integration test already tests a lot + add tests for clip tables

Inverse_clip will be done in a different PR.


hydromt_wflow/components/staticmaps.py
```diff
@@ -249,6 +254,116 @@ def drop_vars(self, names: list[str], errors: str = "raise"):
...
+                "Kind 'geom' for the region is not recommended as it can lead "
+                "to mistakes in the catchment delineation. Use carefully."
+            )
+            geom = hydromt.processes.region.parse_region_geom(region)
+        else:
+            raise ValueError(
+                f"wflow region kind not understood or supported: {kind}. "
+                "Use 'basin', 'subbasin', 'bbox' or 'geom'."
+            )
+
```
Tjalling-dejong, line 332: This part is still uncovered by tests. Could you add a test for these cases as well?

hydromt_wflow/components/utils.py
Tjalling-dejong, line 1: test_equal_grid_data has still many lines of uncovered code.

hydromt_wflow/wflow_sediment.py
Tjalling-dejong, line 1: clip method for wflow sediment is still untested.

Tjalling-dejong (commented):
Nice work! You're nearly there. I just added some comments on test coverage and a typo. I was also wondering why we need duplicate data in the linux64 folder.  I understand that it has to do with testing on different OSs, but shoudnt we then test in a different way. I'm just not liking the repository bloating with the data, especially duplicate data.

Tjalling-dejong (approved):
LGTM 👍

PR #571, opened 2025-09-25
Fix navigation bar layout

deltamarnix:
Version switcher is in front, just like in the theme docs itself: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html

## Issue addressed
Fixes part of #351 with the topbar layout

## Explanation
Moved the version switcher to the center, and put the alignment on left.

image

## Checklist


JoostBuitink:
"Drive by comment" ;) Would it be possible to adjust it so the top bar takes only one row of space? So similar to the current stable docs (see below). Perhaps requires the tabs to be renamed/organized, perhaps also using "more" option, similar to hydromt-core?

image

PR #587, opened 2025-10-01
Fix/staticmap coordinates bug

Tjalling-dejong:
## Issue addressed
Generating a new model would introduce mask and meta_subgrid_outlet_idx coordinates in the staticmaps. This was also not picked up by test_equal_grid_data.  

## Explanation
I improved test_equal_grid_data and also added an extra check on the coordinates. Furthermore, I removed the mask coordinate from the staticmaps when writing and set meta_subgrid_outlet_idx as data variable instead as a coordinate

## Checklist

### DIFF
--- a/.github/workflows/compare-benchmark.yml
+++ b/.github/workflows/compare-benchmark.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Setup pixi env
-        uses: prefix-dev/setup-pixi@v0.9.0
+        uses: prefix-dev/setup-pixi@v0.9.1
         with:
           pixi-version: "v0.45.0"
           environments: test-full-py311
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,7 +45,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Setup pixi env
-        uses: prefix-dev/setup-pixi@v0.9.0
+        uses: prefix-dev/setup-pixi@v0.9.1
         with:
           pixi-version: "v0.47.0"
           environments: docs
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up pixi
-        uses: prefix-dev/setup-pixi@v0.9.0
+        uses: prefix-dev/setup-pixi@v0.9.1
         with:
           run-install: false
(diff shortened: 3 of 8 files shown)


hydromt_wflow/components/staticmaps.py
```diff
@@ -125,6 +125,9 @@ def write(
         # Check for input dir
         p_input = Path(self.model.config.get_value("dir_input", fallback=""), p)
 
+        # Clean up data
+        self.data.drop_vars(["mask"], errors="ignore")
```
deltamarnix, line 129: @JoostBuitink Is it okay to drop the mask variable in the staticmaps during the write? Does Wflow use it in any way?

deltamarnix, line 129: @JoostBuitink Is it okay to drop `mask` from the staticmaps? Is that variable used?

hydromt_wflow/wflow_base.py
```diff
@@ -427,6 +427,11 @@ def setup_basemaps(
         # rename idx_out coords
         if "idx_out" in ds_base:
             ds_base = ds_base.rename({"idx_out": "meta_subgrid_outlet_idx"})
+            # Set meta_subgrid as a data variable instead of coordinate
+            da_outlet = ds_base["meta_subgrid_outlet_idx"]
+            ds_base = ds_base.drop_vars("meta_subgrid_outlet_idx")
+            ds_base["meta_subgrid_outlet_idx"] = da_outlet
+
```
deltamarnix, line 434: Copilot suggests to take a look at [`reset_coords`](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.reset_coords.html). Does that do what you expect? Might save you some lines of code.

```suggestion
            # Set meta_subgrid as a data variable instead of coordinate
            ds_base = ds_base.reset_coords("meta_subgrid_outlet_idx", drop=False)
```

deltamarnix (changes requested):
Is it as simple as removing the mask variable? Can you verify that it is never used?

hydromt_wflow/components/staticmaps.py
JoostBuitink: As far as I am aware, the mask variable is not used nor required by wflow. I do see the "mask" layer being used several times in the hydromt_wflow code, however, so perhaps good to check with @hboisgon if this layer has any actual use.

JoostBuitink: See my comment to the comment above!

PR #613, opened 2025-10-21
Remove passthrough functions on models

### LINKED ISSUE #602: Remove set_grid, write_grid, read_grid, etc. from WflowModel
Since there is no limitation in the names of functions, we were considering to rename set_grid, read_grid, and write_grid, to resp. set_staticmaps, read_staticmaps, and write_staticmaps.

It's a simple rename, but requires the user to make some changes in their yaml files.

aslowwriter:
## Issue addressed
Fixes #602 

## Explanation
 The verb methods on the model class have just been passthrough functions for compatibility for a while, and we decided they are not necessary anymore, so this PR removes them. 

## Checklist

### DIFF
--- a/hydromt_wflow/wflow_base.py
+++ b/hydromt_wflow/wflow_base.py
@@ -1748,19 +1748,14 @@ one variable and variables list is not provided."
             return
         self.write_data_catalog()
         _ = self.config.data  # try to read default if not yet set
-        if "staticmaps" in self.components:
-            self.staticmaps.write(filename=grid_filename)
-            self.staticmaps.write_region(
-                filename=str(Path(geoms_folder) / "region.geojson"), to_wgs84=True
-            )
-        if "geoms" in self.components:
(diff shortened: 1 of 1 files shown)


hydromt_wflow/wflow_base.py
```diff
@@ -1750,21 +1749,21 @@ def write(
...
         if "tables" in self.components:
-            self.write_tables()
+            self.tables.write()
         if "states" in self.components:
-            self.write_states(filename=states_filename)
+            self.states.write(filename=states_filename)
 
         # Write the config last as variables can get set in other write methods
-        self.write_config(filename=config_filename)
+        self.config.write(filename=config_filename)
```
LuukBlom: Can you give this the same format as the `WflowBaseModel.read` function? 
I think its weird we do these if checks, since the components are hardcoded to be available in `WflowBaseModel.__init__`

savente93: sure

PR #630, opened 2025-10-27
Only load reservoir_sh and reservoir_hq tables

deltamarnix:
## Explanation
When loading the tables, it looked for all *.csv files and was not specific in reservoir_{sh,hq}_*.csv.
This happens when reservoir_accuracy.csv is written next to the staticmaps folder.
Bug was introduced in combination of #582 and #598

## Checklist


JoostBuitink:
Very quick "drive-by-comment": note that for v0.x models these tables were called "lake_{sh,hq}_xxx.csv". For the model conversion code, these files still need to be read, so just be aware to also update that part of the codebase.

PR #642, opened 2025-11-05
Changed cold start flag to true in model conversion code

### LINKED ISSUE #633: Change cold start flag to `true` in model conversion code
### Kind of request

None

### Enhancement Description

When converting v0.x model, check if `reinit` (v1.0 name is `cold_start__flag`) is set to `false` . If so, reset the value back to `true` (since converting states is not supported, and doesn't have to be supported), and log this as a warning message to ensure the user is aware of this modification. 

### Use case

_No response_

### Additional Context

_No response_

Tjalling-dejong:
Added a check in _convert_to_wflow_v1 that checks if reinit is set to false and then logs a warning that this flag will be set to true because reinit=false is no longer supported in wflow v1. 

## Issue addressed
Fixes #633 

## Explanation

## Checklist

### DIFF
--- a/hydromt_wflow/version_upgrade.py
+++ b/hydromt_wflow/version_upgrade.py
@@ -132,8 +132,10 @@ def _convert_to_wflow_v1(
     """
     if config["model"].get("reinit") == False:
         logger.warning(
-            "The 'reinit' option set to False is no longer supported in Wflow v1. "
-            "It will be converted to 'cold_start__flag = True'."
+            "Converting states is not supported by this conversion code, therefore the "
+            "reinit option (new name cold_start__flag) is adjusted to True. If model "
+            "states are required, please produce new states using a Wflow simulation or"
+            " use the setup_cold_states function."
         )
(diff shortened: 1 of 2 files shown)


hydromt_wflow/version_upgrade.py
```diff
@@ -130,6 +130,12 @@ def _convert_to_wflow_v1(
     config_out: dict
         The converted config.
     """
+    if config["model"].get("reinit") == False:
+        logger.warning(
+            "The 'reinit' option set to False is no longer supported in Wflow v1. "
+            "It will be converted to 'cold_start__flag = True'."
```
JoostBuitink: I would change this message into something like: "Converting states is not supported by this conversion code, therefore the `reinit` option (new name `cold_start__flag`) is adjusted to `True`. If model states are required, please produce new states using a Wflow simulation or use the `setup_cold_states` function."

hboisgon (approved):
LGTM!

PR #658, opened 2025-11-11
Remove non-reachable docs

deltamarnix:
Removed the following folders that are not reachable in the version switcher: preview, v1.0.0rc1, vf{major}.{minor}.{patch}{suffix}.
Removed versions.json, because it is not used. Instead we use switcher.json

These folders are not relevant, as they are not part of the switcher.json file.

## Checklist


savente93 (approved):
Yeah I think this is okay. thanks!

PR #720, opened 2026-03-17
fix: setup_temp_pet_forcing - update docstring and add warning

### LINKED ISSUE #718: No validation when penman-monteith_ requested with sub-daily forcing

### Reproducible Example

--

### Current behaviour

When pet_method='penman-monteith_rh_simple' or 'penman-monteith_tdew' is used with a sub-daily timestepped model/forcing, setup_temp_pet_forcing requests temp_min and temp_max from the data catalog and passes them to meteo.pet() without any check on the source timestep:
`
  elif pet_method == "penman-monteith_rh_simple":
      variables += ["temp_min", "temp_max", "wind", "rh", "kin", "press_msl"]
`
At sub-daily resolution, temp_min and temp_max are not meaningful daily extremes.. they are instantaneous temperatures at each timestep, nearly identical to temp. The PM formula uses these to compute saturation vapour pressure slope Δ = f(Tmax) - f(Tmin) and the vapour pressure deficit term (es_max - es_min). When Tmax ≈ Tmin ≈ Tmean, both terms collapse, significantly underestimating PET. No warning or error is raised.

### Desired behaviour

After loading ds, check the source timestep against the method requirement. If sub-daily and a penman-monteith_* method is requested, raise a warning:

  if "penman-monteith" in pet_method:
      source_freq = pd.infer_freq(ds.time.values[:3])
      if source_freq is not None and pd.to_timedelta("1D") >pd.tseries.frequencies.to_offset(source_freq):
          warnings.warn(
              f"pet_method='{pet_method}' requires daily temp_min/temp_max. "
              f"Source '{temp_pet_fn}' appears to be sub-daily (inferred freq: {source_freq}). "       
              "Consider pet_method='makkink' or 'debruin' for sub-daily inputs.",
          )

### Additional context

The docstring lists required variables for PM as temp_min [°C], temp_max [°C] but does not state that these must be daily Tmin/Tmax. Makkink (or debruin) is not called out as the recommended alternative for sub-daily inputs anywhere in the docstring.

LuukBlom:
## Issue addressed
Fixes #718 

## Explanation
Added the warning and updated docstring as described in the issue.

## Checklist

### DIFF
--- a/hydromt_wflow/wflow_sbm.py
+++ b/hydromt_wflow/wflow_sbm.py
@@ -3646,14 +3646,14 @@ using 'variable' argument."
                 f"Unknown pet_method '{pet_method}'. Select from "
                 f"{list(method_variables.keys())}."
             )
-        _shared_kwargs = dict(
-            data_like=temp_pet_fn,
-            geom=self.region,
-            buffer=1,
-            time_range=(starttime, endtime),
-            variables=variables,
-            single_var_as_array=False,
-        )
+        _shared_kwargs = {
+            "data_like": temp_pet_fn,
+            "geom": self.region,
+            "buffer": 1,
+            "time_range": (starttime, endtime),
+            "variables": variables,
+            "single_var_as_array": False,
+        }
         # Fetch the required (non-wind) variables
         ds = self.data_catalog.get_rasterdataset(**_shared_kwargs)
 


hydromt_wflow/wflow_sbm.py
```diff
@@ -3523,28 +3523,38 @@ def setup_temp_pet_forcing(
...
+          not rely on daily Tmin/Tmax extremes.
 
-        Depending on the methods, `temp_pet_fn` should contain temperature 'temp' [°C],
+        Depending on the method, `temp_pet_fn` should contain temperature 'temp' [°C],
         pressure 'press_msl' [hPa], incoming shortwave radiation 'kin' [W/m2], outgoing
         shortwave radiation 'kout' [W/m2], wind speed 'wind' [m/s], relative humidity
-        'rh' [%], dew point temperature 'temp_dew' [°C], wind speed either total 'wind'
-        or the U- 'wind10_u' [m/s] and V- 'wind10_v' components [m/s].
+        'rh' [%], dew point temperature 'temp_dew' [°C], and/or U- 'wind10_u' and V-
+        'wind10_v' wind components [m/s].
```
hboisgon: Why the change? ``wind`` is also still supported right?

hydromt_wflow/wflow_sbm.py
```diff
@@ -3555,56 +3562,55 @@ def setup_temp_pet_forcing(
...
+                'press_msl' [hPa], 'kin' [W/m2], 'kout' [W/m2]
 
-            * Required variables for De Bruin reference evapotranspiration: \
-'temp' [°C], 'press_msl' [hPa], 'kin' [W/m2], 'kout' [W/m2]
+            * Required variables for 'makkink':
+                'press_msl' [hPa], 'kin' [W/m2]
 
-            * Required variables for Makkink reference evapotranspiration: \
-'temp' [°C], 'press_msl' [hPa], 'kin'[W/m2]
+            * Required variables for 'penman-monteith_rh_simple':
```
hboisgon: See my other comments on ``wind``. For both penman methods, users can use either ``wind`` or (``wind10_u`` + ``wind10_v``)

hboisgon:
> I implemented a solution for the wind variables: we collect all variables for the ds and then manually filter, is this an acceptable/good approach?)
> 
> I write some tests (with AI) to lock in the error behaviour, I wanted to write some tests while Mocking as much as possible.
> 
> I like the tests, in total they run in 1.3sec so I think this should be good to add, but they are quite different from the existing tests, so I didnt want to put them in `test_model_methods.py`.
> 
> This is what I see as unit tests that should be part of coverage, any test where we actually use real data / datacatalog stuff is an integration test imo.
> 
> @hboisgon
> 
> * do you agree these tests should be added
> * do you agree with the location / formatting

Test location and content look good thanks :)

For the variables selection, ERA5 especially from EarthDataHub catalog has approx 130+ variables so performance might considerably drop....
I think I would still do the old implementation and keep the wind out of the list and read it in a second step. For wind, if you do get_rasterdataset(variable=["wind"], handle_nodata=NoDataStrategy.IGNORE) do you get an error or None? I think ideally we should get None and then we can try on variables = ["wind10_u", "wind10_v"] but if an error is resturned then open an issue for core for later and for now a try with variables = wind and except with u/v components?
Also error messages in the log coming from ``exec_no_data_strategy`` are super criptic, in hydromt_wflow code I would just use a classic raise ValueError.

PR #728, opened 2026-03-23
Support for rasterio 1.5

### LINKED ISSUE #694: resolve problems with rasterio 1.5

### Reproducible Example

remove the version limitation of `rasterio` in `pyproject.toml`
run `pixi run build-example-models`

### Current behaviour

When writing the first model, this cryptic error is raised:

```txt
2026-01-06 13:15:34,271 - hydromt.hydromt_wflow.components.config - config - INFO - Writing model config to C:/Users/blom_lk/Projects/hydromt/hydromt_wflow/examples/wflow_piave_clip/wflow_sbm.toml.
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
PS C:\Users\blom_lk\Projects\hydromt\hydromt_wflow>
```

### Desired behaviour

no error 

### Additional context

rasterio 1.5 also breaks tests when comparing to the existing example models for linux and windows with python versions > 3.11

See workflow run: https://github.com/Deltares/hydromt_wflow/actions/runs/20745980777

LuukBlom:
## Issue addressed
Fixes #694 

## Explanation

This PR does the updates required for rasterio 1.5, which changes something in interpolation / mode selection, that produce different results than before. (see the `test_setup_lai`)
- Removed limit for rasterio 1.5

This PR also resolves the error described in https://github.com/Deltares/hydromt_wflow/issues/694.
- Added safe_excepthook to toplevel ``__init__.py`` to prevent the sys.excepthook error due to rasterio installing a broken hook

This PR also has quite some line changes in the tests, but these are not changing any behaviour:
- Refactored static Path-like singletons into session scoped fixtures (now they no longer need to be imported in other test files, since the fixtures are made available by pytest.)
- Added plotting helper functions that create plots when `_compare_wflow_models` fails.
- Updated handling of Path-like variables: all usage of the builtin `os` module is now replaced by the recommended `pathlib`

## Files to review
- `hydromt_wflow/__init__.py` -- this contains the fix for #694
- `tests/conftest.py` -- contains the fixture updates
- `tests/test_model_class.py` -- contains the plotting functions
- `tests/test_model_methods.py::test_setup_lai` -- conditionally assert depending on the rasterio version 
- All other changes are trivial changes / updates that do not change behaviour.


LuukBlom:
Regenerating the example models with `pixi run build-example-models` fixes 2 / 3 errors: 
- `tests/test_model_class.py::test_model_build[wflow]` 
- `tests/test_model_class.py::test_model_build[wflow_sediment]`
But not `test_setup_lai`

@hboisgon are the `test_model_build` errors ones that should be fixed by regenerating the models or is this a code issue somewhere? 

Also, what can you say about the `test_setup_lai` error? Its quite difficult for me to guess what is wrong.

hydromt_wflow/components/utils.py
```diff
@@ -37,13 +38,17 @@ def _relpath(
...
+            if sys.version_info >= (3, 12):
+                _path = _path.relative_to(root, walk_up=True)
+            else:
+                # remove when Python 3.11 support is dropped
+                _path = Path(os.path.relpath(_path, root))
     except ValueError:
-        pass  # `value` path is not relative to root
-    return value.as_posix()
+        pass  # `_path` path is not relative to root
+    return _path.as_posix()
```
deltamarnix, line 51: I wonder how much this function is still worth. Why is `value` `Any`? Shouldn't users just work with pathlib anyway?

I'm okay with it for now, as we don't want to make this PR any larger. But consider if we should create a ticket that removes this function.

hydromt_wflow/workflows/river.py
```diff
@@ -611,8 +610,8 @@ def _discharge(ds_like, flwdir, da_precip, da_climate):
     # read clim classes and regression parameters from data dir
     precip_fn = da_precip.name
     climate_fn = da_climate.name
-    fn_regr = join(DATADIR, "rivwth", f"regr_{precip_fn}.csv")
-    fn_clim = join(DATADIR, "rivwth", f"{climate_fn}.csv")
+    fn_regr = str(DATADIR / "rivwth" / f"regr_{precip_fn}.csv")
+    fn_clim = str(DATADIR / "rivwth" / f"{climate_fn}.csv")
```
deltamarnix: I don't think there is a need for `str()` here, because pandas accepts pathlike objects.

hydromt_wflow/__init__.py
```diff
@@ -1,5 +1,33 @@
...
+        _traceback.print_exception(exc_type, exc_value, exc_tb, file=_sys.stderr)
+        try:
+            _original_excepthook(exc_type, exc_value, exc_tb)
+        except Exception:
+            pass  # suppress the broken hook
+
+    _sys.excepthook = safe_excepthook
+
+except Exception:
+    print(
```
deltamarnix: If this happens, what should the user do? Are there workarounds? It would be good to give the user some more explanation.

hydromt_wflow/__init__.py
```diff
@@ -1,5 +1,33 @@
 """HydroMT plugin for wflow models."""
 
+try:
+    # Rasterio 1.5 installs a broken sys.excepthook that recurses infinitely
+    # on interpreter shutdown. Reset it to a safe version that prints the exception
+    # and then calls the original hook wrapped in a try-except.
+    import sys as _sys
+    import traceback as _traceback
+
+    import rasterio as _rasterio
```
deltamarnix, line 12: Is this import statement the moment that the excepthook is placed? You code is now set on all versions of rasterio, not only 1.5. So what happens with other versions? Will you see the same error message twice?

hydromt_wflow/__init__.py
```diff
@@ -1,5 +1,33 @@
...
+
+    _original_excepthook = _sys.excepthook
+
+    def safe_excepthook(exc_type, exc_value, exc_tb):
+        print("REAL EXCEPTION:", file=_sys.stderr)
+        _traceback.print_exception(exc_type, exc_value, exc_tb, file=_sys.stderr)
+        try:
+            _original_excepthook(exc_type, exc_value, exc_tb)
+        except Exception:
+            pass  # suppress the broken hook
```
deltamarnix: To me it would make sense to put the `print_exception` inside of the `except`, so that you don't see the exception twice. That way there is also no need to print `REAL EXCEPTION`.


```suggestion
        try:
            _original_excepthook(exc_type, exc_value, exc_tb)
        except Exception:
             _traceback.print_exception(exc_type, exc_value, exc_tb, file=_sys.stderr)
```

hydromt_wflow/wflow_base.py
```diff
@@ -1029,7 +1027,7 @@ def setup_lulcmaps_from_vector(
         )
 
         if save_raster_lulc:
-            lulc_out = join(self.root.path, "maps", "landuse_raster.tif")
+            lulc_out = str(self.root.path / "maps" / "landuse_raster.tif")
```
deltamarnix: Can this end result be a `Path`?

tests/test_utils.py
```diff
@@ -162,7 +160,9 @@ def test_convert_to_wflow_v1_sediment():
     )
 
     # Check with a test config
-    config_fn_v1 = join(TESTDATADIR, "wflow_v0x", "sediment", "wflow_sediment_v1.toml")
+    config_fn_v1 = str(
```
deltamarnix, line 163: It was the easiest to replace every join with a str, but I wonder if WflowSedimentModel.config_filename can simply be of type Path.

deltamarnix (changes requested):
I have some questions regarding the excepthook that you implemented. The pathlib changes look good to me.

hydromt_wflow/components/utils.py
LuukBlom: I agree. Will make an issue for dropping python 3.11 and this should be part of it.
https://github.com/Deltares/hydromt_wflow/issues/738

hydromt_wflow/__init__.py
LuukBlom: Yes, the first time rasterio is imported, is when the broken hook is installed. 

In other versions of rasterio, there is no error on interpreter shutdown.
When there is no error, there is no output in the terminal.

tests/test_utils.py
LuukBlom: True, in almost all cases I think they are interchangeable and I would prefer to always use pathlib over str / os.path.

However, a lot of the type hints for functions expect strings and not Paths.
So, updating that requires changing type hints / function signatures (which possibly requires more changes in the functions themselves to account for the type variance).
Which I didnt want to do to keep the already large amount of changes as simple as possible.

I can still do it here to have these updates in one PR, but can also make an issue for it.

deltamarnix (approved):
I think we're there. Pity that we need this workaround. I would like to see an issue on rasterio that addresses this and an issue in hydromt_wflow so that we can track the problem and hopefully remove the workaround again soon.

deltamarnix:
I accept the Sonarcloud code analysis issue.

PR #744, opened 2026-04-30
fix: add `river_depth_kwargs` as optional arg to `setup_rivers`

### LINKED ISSUE #710: extra params in `setup_rivers`
Support kwargs for the different `rivdph` method params in `setup_rivers`.
In my case, I wanted to modify the default values of `hc` and `hp` in the powlaw method.

If you don't want to,  consider modifying the docs

"_The river depth can be directly derived from river_geom_fn property or calculated using the rivdph_method, by default powlaw: h = hc*Qbf**hp, which is based on qbankfull discharge from the nearest river segment in river_geom_fn and takes optional arguments for the hc (default = 0.27) and hp (default = 0.30) parameters._"

LuukBlom:
## Issue addressed
Fixes #710 

## Explanation

`hydromt.model.processes.river_depth` already exposes `kwargs` that can be used like this.

### DIFF
--- a/hydromt_wflow/wflow_base.py
+++ b/hydromt_wflow/wflow_base.py
@@ -478,6 +478,7 @@ skipping adding gauge specific outputs to the toml."
         min_rivwth: float = 30,
         rivdph_method: str | None = "powlaw",
         min_rivdph: float = 1,
+        river_depth_kwargs: dict | None = None,
         output_names: dict = {
             "river_location__mask": "river_mask",
             "river__length": "river_length",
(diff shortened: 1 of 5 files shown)


evetion (approved):
If there is an existing test for this workflow (even a high-level integration one), it would be good to provide some river_depth_kwargs.

hydromt_wflow/wflow_base.py
deltamarnix, line 471:
```suggestion
        self,
        *,
```

I see that SonarQube doesn't like so many arguments. I can sort of understand that. I would like to propose that we at least make the arguments keyword arguments, instead of positional.

deltamarnix (changes requested):

PR #757, opened 2026-06-01
fixes for hydromt v1.4.0

LuukBlom:
## Issue addressed
Fixes #<issue number>

## Explanation

## Checklist


tests/components/test_staticmaps_component.py
deltamarnix, line 228: tmp_path should exist on fixture usage.

LuukBlom: I missed that `ModelRoot._cleanup()` now deletes the model dir if its empty.
This is called when we overwrite the `.root` attribute of the model and trigger the `__del__` of the ModelRoot.

tests/components/test_geoms_component.py
```diff
@@ -133,9 +133,12 @@ def test_wflow_geoms_component_set(
     # Check that the geometry matches
     gdf_read = new_comp.get("test_geom")
     assert gdf_read is not None
-    assert gdf_read.geom_equals_exact(mock_geometry, tolerance=1e-7).all(), (
-        "Read geometry does not match the original geometry."
-    )
+
+    assert (
```
evetion, line 137: Why do we need to normalize geometries now, and why did the tolerance drop?

evetion (approved):
Best to update the PR description to link to the other PRs that incurred the changes here, otherwise it's hard to review.

tests/components/test_geoms_component.py
LuukBlom: Now needed because in core we set the precision, which re-ordered the coordinates

PR #765, opened 2026-06-11
feat: update config parameters and add upgrade to v1.1

LuukBlom:
## Issue addressed
Fixes #727 

## Explanation

When working on this, we realized that the version upgrade and internal naming system were interwoven, making it difficult to extend it for multiple chained version upgrades. 

**How naming used to work:** 

From a dict that contains the variable names for wflow for v0 and v1.0, as well as the staticmap and hydromt names for that variable, create 2 mappings: `wflow_to_staticmap` and `hydromt_to_staticmap`. These mappings are then used to write the staticmaps and config correctly to disk such that wflow.jl understands it. 

In the middle of creating the mappings, the upgrades were also done, making it quite difficult to add a third supported upgrade version.

**How naming now works:** 
Naming should expect to only have to read in the latest version and not have to do upgrades on the fly, that should be an explicit step taken by the user. 

So, now we have a dict for the latest version only: staticmap_name to wflow_name. 
Each entry in the dict is a parameter. Each parameter has:
- staticmap name (key in the dict, also the name of the in-memory staticmap layer)
- wflow_name (value in the dict, also the name of the on-disk staticmap layer)
- hydromt_name (optional), mainly used as standard name in data that is stored at deltares by hydromt datacatalogs. 

**How upgrading used to work:** 

A Model had to be initialized to run the upgrade. Meaning the newer models always have to be able to read in older models without errors (this is what the stuff in naming was doing, to then update stuff in memory and write it to disk. 

**How upgrading now works:** 

This is now an explicit function in the version upgrade module that takes in a model root to upgrade: `hydromt_wflow.version_upgrade.upgrade_model(model_root: Path, ...)`

## Summary

- Renamed config parameters as described in the issue
- Refactored naming and version upgrade; these are now separate systems. 
- Implemented the upgrade from v1.0 to v1.1
- Rebuild all example models (windows/linux) for v1.1
- Added `pixi run ` to all cli commands in examples to get them to run (how did this work before? would like to revert these changes)
- Refactored tests for upgrading v0 to v1
- Added tests for upgrading v1 to v1.1
- Cleanup and organize the data required for this testing. (`tests/data/upgrade` has all data for testing. `examples/data/wflow_upgrade` has the files for the upgrade example.)

# For reviewers

There are many many files changed, but not that many significant ones. Most changes are updating the example models / other data in the repo. 

These are the changes from most relevant/in need of review, to least relevant / interesting:

### version upgrade
- `version_upgrade.py & wflow_sbm.py & wflow_sediment.py`: new system to upgrade to latest version
- `tests/test_version_upgrade.py`: tests for the upgrade system (isolated tests for each upgrade step, aggregate test for v0 -> latest as well)
- `tests/data/upgrade`: organized models + data for testing the upgrades (no longer uses example models.)

### naming
- `hydromt_wflow/workflows & naming.py`: config parameter renames according to the issue.`

### other
- `.github/workflows/system_test.yml`: now we use `Wflow.jl@master` instead of `Wflow.jl@release/v1.0`. TODO!
- `hydromt_wflow/data`: added wflow_version = "1.1" to default config
- `tests/components`: updates for config component, "wflow_version" now in config.
- `examples/*.ipynb`: updates to cli commands that were broken. 
- `examples/`: rebuilt example models on windows and linux, updated config tomls.
- `docs/`: renamed `upgrade_to_wflow_v1` to `upgrade_to_latest`

# TODO 

- [ ] discuss the upgrade design with @deltamarnix @evetion 
- [ ] validate parameters updates are correct
- [ ] fix system test errors & point to `release/v1.1` in wflow.jl once it exists


examples/build_model.ipynb
```diff
@@ -21,7 +21,7 @@
    "id": "2",
    "metadata": {},
    "source": [
-    "All lines in this notebook which start with `!` are executed from the command line. Within the notebook environment the logging messages are shown after completion. You can also copy these lines and paste these in your shell to get more direct feedback."
+    "All lines in this notebook which start with `!pixi run ` are executed from the command line. Within the notebook environment the logging messages are shown after completion. You can also copy these lines and paste these in your shell to get more direct feedback."
```
deltamarnix: I think you can revert this line. It was correct that anything with `!` is run on the command line.

examples/build_model.ipynb
```diff
@@ -49,7 +49,7 @@
    "source": [
     "# this should return \"wflow_sbm, wflow_sediment\"\n",
     "# as well as the generic HydroMT model \"model\"\n",
-    "!hydromt --models"
+    "!pixi run hydromt --models"
```
deltamarnix: Adding `pixi run` would make it impossible for uv users to run the examples locally. Not recommended.

hydromt_wflow/wflow_sbm.py
```diff
@@ -4078,3 +4088,65 @@ def upgrade_to_v1_wflow(self):
...
+            )
+            version = Version("0.0")
+        else:
+            version = Version(str(version_str))
+
+        if version >= WFLOW_LATEST_VERSION:
+            logger.info("Model is already at the latest version, no upgrade needed.")
+            return
+
+        if version < Version("1.0"):
```
deltamarnix: Instead of if-statements, can you make some cool dictionary or array that will automatically execute with a for loop? For example:

```py
upgrades = SortedDict({
    "1.0": self._upgrade_v0_to_v1,
    "1.1": self._update_v1_to_v1_1
})

for v, upgrade in upgrades:
    if version < Version(v):
        upgrade()
        version = self.config.wflow_version
```

Would something like that work? Making it easier to add new versions when we have over 10 of these functions.

hydromt_wflow/wflow_sbm.py
```diff
@@ -4078,3 +4088,65 @@ def upgrade_to_v1_wflow(self):
                 self.config.set(option, config_opt[option])
         # also update tables
         upgrade_lake_tables_to_reservoir_tables_v1(self.tables)
+
+    def _upgrade_v1_to_v1_1(self):
```
deltamarnix: I think the upgrade functions can move to another file where it makes more sense. This python file is already over 4000 lines. Can we move it to workflows of something alike?

hydromt_wflow/wflow_sbm.py
```diff
@@ -4078,3 +4088,77 @@ def upgrade_to_v1_wflow(self):
...
+                f"but found '{version}'. Please run _upgrade_v0_to_v1() first."
+            )
+
+    @hydromt_step
+    def upgrade_to_v1_wflow(self):
+        """Upgrade the model to Wflow v1.0 format."""
+        warnings.warn(
+            "`upgrade_to_v1_wflow()` is deprecated and will be removed in a future "
+            "release, use `upgrade_to_latest()` instead.",
+            DeprecationWarning,
```
deltamarnix: You could also add `@deprecated`. Would that have the same result?

hydromt_wflow/wflow_sediment.py
```diff
@@ -1076,6 +1078,64 @@ def upgrade_to_v1_wflow(
...
+        Applies all necessary upgrade steps in order based on the ``wflow_version``
+        key in the config. If absent, the model is assumed to be pre-v1.0 and all
+        upgrade steps are applied.
+
+        This function should be followed by write() to write all upgraded components
+        to disk.
+        """
+        version_str = self.config.data.get("wflow_version")
+        if version_str is None:
+            logger.warning(
```
deltamarnix: This is incorrect. wflow 1.0 models all already created out in the world without `wflow_version` written in that config. Unfortunately, you will have to find another way to find out if the model was 1.0. For example, by checking a certain config variable that only exists in 1.0 and not in 0.x

tests/data/upgrade/sbm/v0x/wflow_sbm.toml
```diff
@@ -13,7 +13,7 @@ path_output = "outstate/outstates.nc"
 
 [input]
 path_forcing = "inmaps.nc"
-path_static = "staticmaps/staticmaps_v0x.nc"
```
deltamarnix, line 16: I didn't expect this file to be upgraded, since it is a specific v0x file.

deltamarnix (changes requested):
I know you're still working on this, but I already want to give some feedback on what I see happening.

examples/build_model.ipynb
LuukBlom: Yes, I would like to revert these too, but the docs were throwing errors caused by these commands not running.
Will investigate more to hopefully revert

hydromt_wflow/wflow_sbm.py
LuukBlom: I like moving it as well. but I think moving this all to workflows could be difficult, since we need access to the model and components. So they wont be functions like the other workflows that are just data pipelines. they would actually change the models, so perhaps we put them in a different module altogether.

The simplest could be, just take in a `WflowSbmModel` as an argument to these functions, then update the component data and set it again.

But perhaps the cleanest is to move the upgrade code to the components and then call `component.upgrade()` on all of them. This would then clutter the component code, and could be problematic when multiple components need simultaneous upgrading.

hydromt_wflow/wflow_sbm.py
LuukBlom: I think so, will have a look tomorrow

tests/data/upgrade/sbm/v0x/wflow_sbm.toml
LuukBlom: I changed the file structure to have the model types (`sbm/sediment`) and versions (`v0x/v1_0/v1_1`) be the directories, then each model has the default names for the files.

Each model dir has its own staticmaps & config.toml that we can use in tests for assertions.
previous impl was difficult to extend nicely for more versions.

LuukBlom:
Just talked to @JoostBuitink about this, and there is a decision required form the wflow.jl team about the config parameters. 
It is likely that `land_elevation` will be split into 2 parameters: `floodplain_elevation` and `land_elevation`. The land_elevation one should be added in setup_basemaps. 

He will discuss with the team and come back here with the names of the newly added required config parameters.

vers-w:
> It is likely that `land_elevation` will be split into 2 parameters: `floodplain_elevation` and `land_elevation`. The land_elevation one should be added in setup_basemaps.

Note that it should be `land_surface__elevation`. Just wondering, why do we need the extra parameter `floodplain_elevation`?

LuukBlom:
> Note that it should be `land_surface__elevation`. Just wondering, why do we need the extra parameter `floodplain_elevation`?

Why the second param is needed, is a question for @JoostBuitink 

Why we use different names:
wflow.jl parameter names can be quite long and strangely formatted (`__` / `~` / etc), so a naming system was developed that we use in hydromt_wflow to map to and from wflow.jl variables, without having to use them directly. (the system could be much better, but this is how its been set up)

The mappings are defined in naming.py.

For this parameter:

- `elevtn` is the hydromt name.
- `land_elevation` is the hydromt_wflow staticmap layer name.
- `land_surface__elevation` is the Wflow.jl variable name (v1.1).

Most setup methods call `model._update_naming()`, which uses these mappings to translate between hydromt_wflow names and Wflow.jl names.

So, even though hydromt_wflow internally uses `land_elevation`, the written staticmaps.nc will have the `land_surface__elevation` layer

vers-w:
> Why the second param is needed, is a question for @JoostBuitink

Thanks for the explanation Luuk! I indeed thought this was about Wflow.jl parameter names.

Briefly discussed this today with Joost, and the second parameter is required: `land_elevation` refers to the land surface elevation (e.g. DEM) while `land_surface_water_flow__ground_elevation` (Wflow name) refers to the land elevation for local inertial overland flow routing (2D). The latter makes use of conditioned DEM to D4 flow directions (I think `hydrodem` in hydrom_wflow).

```
    "land_elevation": {
        "wflow_v0": "lateral.land.elevation",
        "wflow_v1": "land_surface_water_flow__ground_elevation",
        "hydromt_name": "elevtn",
    },
```

JoostBuitink:
I am fairly confident that something goes wrong in the 2D floodplains function: as @vers-w mentioned, there is a D4-conditioned DEM that should be set to the `land_surface_water_flow__ground_elevation` parameter, but that currently gets the DEM. I'll do a quick test after this and create a new issue if it is indeed the case.

This also means that an update to the naming dictionary is preferred, see my proposal below (also tagging @vers-w for his opinion). Please note the comments I put in this as well

```python
    # NOTE: this was "land_elevation" before
    "land_elevation_D4": { 
        "wflow_v0": "lateral.land.elevation",
        "wflow_v1": "land_surface_water_flow__ground_elevation",
        "hydromt_name": "hydrodem_D4",
    },
    "land_elevation": {
        "wflow_v0": "altitude", # NOTE: old (sbm_gwf) models already had this key, although it was stored under [input], and not under [input.vertical] or [input.lateral] like most other keys. So please check if this works
        "wflow_v1": "land_surface__elevation", # this was actually already a field in v1.0, however, it now becomes a hard requirement in v1.1
        "hydromt_name": "elevtn",
    },
```

For getting the "land_surface__elevation", I think this dictionary needs to be updated https://github.com/Deltares/hydromt_wflow/blob/ac1dd879b23439d8af7c89f6257b1ad22db2164f/hydromt_wflow/wflow_base.py#L297-L301

and note these two sections as well (right not elevation is explicitly skipped, this is no longer needed)
https://github.com/Deltares/hydromt_wflow/blob/ac1dd879b23439d8af7c89f6257b1ad22db2164f/hydromt_wflow/wflow_base.py#L443-L446

https://github.com/Deltares/hydromt_wflow/blob/ac1dd879b23439d8af7c89f6257b1ad22db2164f/hydromt_wflow/wflow_base.py#L460-L463

JoostBuitink:
> I am fairly confident that something goes wrong in the 2D floodplains function: as @vers-w mentioned, there is a D4-conditioned DEM that should be set to the `land_surface_water_flow__ground_elevation` parameter, but that currently gets the DEM. I'll do a quick test after this and create a new issue if it is indeed the case.

Decided to directly tackle it in this PR, perhaps easier to do it in one go with the expected change in the naming dictionary as well. I guess that the tests will fail as it will produce a slightly different toml, so that would need a small fix

examples/wflow_piave_subbasin/wflow_sbm_results.toml
```diff
@@ -48,7 +49,7 @@ snowpack_dry_snow__leq_depth = "snow_leq_depth"
 snowpack_liquid_water__depth = "snow_water_depth"
 land_surface_water__instantaneous_volume_flow_rate = "land_instantaneous_q"
 land_surface_water__depth = "land_h"
-subsurface_water__volume_flow_rate = "subsurface_q"
+subsurface_water__instantaneous_volume_flow_rate = "subsurface_q"
```
deltamarnix, line 52: Did you regenerate the examples or did you make the changes by hand?

hydromt_wflow/version_upgrade.py
```diff
@@ -868,3 +880,341 @@ def upgrade_lake_tables_to_reservoir_tables_v1(tables: WflowTablesComponent) ->
...
+    model: WflowSbmModel | WflowSedimentModel
+        The model to upgrade.
+    options: dict, optional
+        A dictionary of options to pass to the upgrade functions. The keys should be
+        strings of the format "from_to" (e.g. "0.8_1.0") and the values should be dicts
+        of options to pass to the corresponding upgrade function.
+
+    Example yml:
+    ```
+        0.8_1.0:
```
deltamarnix: The "from" might not be bothered by the user, because it is always "previous_to_X.X". I would suggest to name them: "to_1.0" and "to_1.1". Or just the version number it will upgrade to. "1.0" and "1.1".

hydromt_wflow/version_upgrade.py
```diff
@@ -868,3 +880,341 @@ def upgrade_lake_tables_to_reservoir_tables_v1(tables: WflowTablesComponent) ->
...
+        "No 'wflow_version' found in config and no v1.0-only keys detected. "
+        "Assuming pre-v1.0 model with v0.8 schema. Please check the config and update "
+        "the version if needed."
+    )
+    return Version("0.8")
+
+
+def _validate_options(
+    options: dict[str, dict[str, dict]] | None, model_name: str
+) -> dict[tuple[Version, Version], dict[str, dict]]:
```
deltamarnix, line 1266:
```suggestion
def _validate_options(
    options: dict[str, Any] | None, model_name: str
) -> dict[tuple[Version, Version], Any]:
```

If you have dicts of dicts of dicts, I'm not sure if we should keep on listing them. Maybe just Any is enough.

hydromt_wflow/version_upgrade.py
```diff
@@ -868,3 +880,341 @@ def upgrade_lake_tables_to_reservoir_tables_v1(tables: WflowTablesComponent) ->
...
+        _from, _to = version_string.split("_")
+        _from_v = Version(_from)
+        _to_v = Version(_to)
+        version_tuple = (_from_v, _to_v)
+        if version_tuple not in _UPGRADES[model_name]:
+            raise ValueError(
+                f"Unknown upgrade versions '{version_tuple}' for model '{model_name}'. "
+                f"Available versions: {list(_UPGRADES[model_name].keys())}."
+            )
+        validated[version_tuple] = opts
```
deltamarnix, line 1294: You know what would be nice?! To use partial functions here.

Instead of return a dictionary with options, you can return a list of functions that need to be executed after one another and the arguments are tucked inside of partial functions.

```py
def get_upgrade_function_train(options, model_name):
    upgrade_function_train = []
    for version_tuple, upgrade_func in _UPGRADES[model_name]:
        # TODO: first sort based on version_tuples?
        upgrade_function_train.append(partial(upgrade_func, **options[version_tuple]))
    return upgrade_function_train

# and later in the code just call them
for f in upgrade_functions:
    f(model)
```

hydromt_wflow/version_upgrade.py
```diff
@@ -868,3 +880,341 @@ def upgrade_lake_tables_to_reservoir_tables_v1(tables: WflowTablesComponent) ->
...
+
+
+def convert_to_wflow_v1_1_sbm(config: dict) -> dict:
+    """Convert the config of a Wflow v1.0 model into a Wflow v1.1 format for SBM."""
+    config_out = copy.deepcopy(config)
+
+    all_vars = {**WFLOW_NAMES, **WFLOW_STATES_NAMES}
+    for variables in all_vars.values():
+        v1 = variables.get("wflow_v1")
+        v1_1 = variables.get("wflow_v1.1")
```
deltamarnix: Wouldn't it make our lives easier if we make separate dictionaries for every upgrade step? I know the 1.0 was rather large, but it's easier to have a small dictionary that only shows how you get to a v1.1, instead of a big dictionary with everything included.

deltamarnix (changes requested):

tests/data/upgrade/sbm/v0x/wflow_sbm.toml
deltamarnix: That sounds good, but why are there some new variables in this file? reservoirlocs, input.vertical.c, csv.column

tests/data/upgrade/sbm/v1_0/wflow_sbm.toml
```diff
@@ -15,7 +16,7 @@ cold_start__flag = true
 glacier__flag = true
 land_routing = "kinematic_wave"
 river_routing = "kinematic_wave"
-kinematic_wave__adaptive_time_step_flag = false
+kinematic_wave__adaptive_time_step_flag = true
```
deltamarnix, line 19: Why did this change from false to true?

tests/data/upgrade/sbm/v1_1/wflow_sbm.toml
```diff
@@ -111,8 +110,24 @@ reservoir_surface__area = "reservoir_area"
 reservoir_water_demand__required_downstream_volume_flow_rate = "reservoir_demand"
 reservoir_water_release_below_spillway__max_volume_flow_rate = "reservoir_max_release"
 reservoir_water__max_volume = "reservoir_max_volume"
+reservoir_water__target_full_volume_fraction = "reservoir_target_full_fraction"
 reservoir_water__target_min_volume_fraction = "reservoir_target_min_fraction"
 
+[input.static.soil_layer_water__brooks_corey_exponent]
```
deltamarnix, line 113: Same as other files. Why did the configs change?

hydromt_wflow/__init__.py
```diff
@@ -31,13 +31,17 @@ def safe_excepthook(exc_type, exc_value, exc_tb):
...
 
-DATA_DIR = Path(__file__).parent / "data"
-
-__all__ = ["WflowBaseModel", "WflowSbmModel", "WflowSedimentModel", "DATA_DIR"]
+__all__ = [
+    "WflowBaseModel",
+    "WflowSbmModel",
+    "WflowSedimentModel",
+    "DATA_DIR",
+    "upgrade_model",
```
deltamarnix: why did you decide to export this one function?

hydromt_wflow/wflow_sbm.py
```diff
@@ -525,11 +519,11 @@ def setup_floodplains(
...
             ).rename(name)
             self.staticmaps.set(ds_out)
-            # Update the bankfull elevation map
+            # Update the bankfull elevation map, and also use this for the ground
+            # elevation
             self.config.set("input.static.river_bank_water__elevation", name)
-            # In this case river_bank_elevation is also used for the ground elevation?
             self.config.set(
-                "input.static.land_surface_water_flow__ground_elevation", elevtn_map
+                "input.static.land_surface_water_flow__ground_elevation", name
```
deltamarnix, line 526: Could you verify that this is correct?

hydromt_wflow/wflow_base.py
```diff
@@ -65,7 +69,7 @@ class WflowBaseModel(Model):
     # TODO supported model version should be filled by the plugins
     # e.g. _MODEL_VERSION = ">=1.0, <1.1
 
-    _DATADIR: Path = utils.DATADIR
+    _DATADIR: Path = utils.DATA_DIR
```
deltamarnix: Looks like a variable that can be removed?

hydromt_wflow/wflow_base.py
```diff
@@ -1714,6 +1716,51 @@ def clip(
...
+        config_filename: str | None = None,
+    ):
+        """Upgrade the model to the latest Wflow.jl version.
+
+        First, copies the model to a new location to keep the original model unchanged.
+        Then, applies all necessary upgrade steps in order based on the
+        ``wflow_version`` key in the config. If absent, the model is assumed to be
+        pre-v1.0 and all upgrade steps are applied.
+
+        This function should be followed by write() to write all upgraded components
```
deltamarnix, line 1742: When this is done in a yaml, does the user need to specify the output_dir, or is that handled by upgrade_model function?

hydromt_wflow/version_upgrade.py
```diff
@@ -861,10 +1152,345 @@ def convert_reservoirs_to_wflow_v1_sediment(
...
+    return config_out
+
+
+# Helper functions for version detection and validation for the upgrade functions.
+def _is_v1_schema(config: dict) -> bool:
+    """Detect the Wflow v1.0 config schema."""
+    # v1.1+
+    if "wflow_version" in config:
+        logger.debug("Found wflow_version, indicating v1 schema.")
+        return True
```
deltamarnix: I'm not sure if I understand this function. The description says that it detects a v1.0 config schema, but it returns true if a "wflow_version" is present, which means it's actually a 1.1 schema. So why does it return true?

hydromt_wflow/version_upgrade.py
```diff
@@ -861,10 +1152,345 @@ def convert_reservoirs_to_wflow_v1_sediment(
...
+    }
+
+    if v1_model_keys.intersection(model_cfg):
+        logger.debug("Found v1 model options.")
+        return True
+
+    # v1 output structure
+    output_cfg = config.get("output", {})
+
+    if "netcdf_grid" in output_cfg or "netcdf_scalar" in output_cfg:
```
deltamarnix, line 1224: If we get all the way here, we expect that all the other stuff was not present, or it was still in v0 format. How do you check that difference? If it is a Frankenstein version, I'm not sure if we should be returning true by now.

hydromt_wflow/version_upgrade.py
deltamarnix, line 613: Would it be possible to start with the original config and move/remove the parts that have changed? It looks like we have to go over all parts of the config, but actually most of the config is in the right shape. There are only changes to take care of.

hydromt_wflow/version_upgrade.py
```diff
@@ -228,12 +676,12 @@ def _update_output_netcdf_grid(wflow_var, var_name):
...
-            key=f"state.{variables['wflow_v0']}", config=config, fallback=None
-        )
-        if name is not None and variables["wflow_v1"] is not None:
-            config_out["state"]["variables"][variables["wflow_v1"]] = name
+    for v0_var, v1_var in v0_to_v1_states.items():
+        if v1_var is None:
+            continue
+        name = get_config(key=f"state.{v0_var}", config=config, fallback=None)
+        if name is not None:
+            config_out["state"]["variables"][v1_var] = name
```
deltamarnix, line 684: Would it be easier to use a toml library that can directly handle variables with periods (.) in the names? Now I see we have to build up a bunch of dictionary entries with empty statements, while we actually just want to move things around.

deltamarnix (changes requested):
I haven't checked the whole migration, but I have the idea that changing the config itself can be simplified by having a starting point, instead of rebuilding it from the ground up.

hydromt_wflow/wflow_base.py
LuukBlom: its uses what is provided, or creates a copy in a tmp dir, and upgrades that.

tests/data/upgrade/sbm/v1_0/wflow_sbm.toml
LuukBlom: In the removed test `test_utils.py::test_convert_to_wflow_v1_sbm_with_exceptions`, this was set to false to make the assert pass, which is also weird imo.

tests/data/upgrade/sbm/v0x/wflow_sbm.toml
LuukBlom: The previous version tested v0 without lakes and v0 with lakes separately, which seems redundant to me, so I merged them. Now we test with one v0 model that has lakes. 

The changes are the lakes that I added from the other v0x.
See [tests/data/wflow_v0x/sbm_with_lake_files/wflow_sbm_v0x.toml](https://github.com/Deltares/hydromt_wflow/blob/main/tests/data/wflow_v0x/sbm_with_lake_files/wflow_sbm_v0x.toml)

examples/wflow_piave_subbasin/wflow_sbm_results.toml
LuukBlom: I regenerated them

hydromt_wflow/version_upgrade.py
LuukBlom: I like being explicit here. I think it communicates to users exactly what we expect and forces users to understand the state and version of their model. 
`to_1.1` or `1.1` could be misinterpreted, for either the whole upgrade chain up until 1.1, or 1.0 to 1.1, or even 1.1 to 1.2 in the future.

tests/data/upgrade/sbm/v1_1/wflow_sbm.toml
LuukBlom: These changes are due to the merge of the previous v0 models in `examples/data/wflow_upgrade` (with / without lakes).

The merged model is at `tests/data/upgrades/sbm/v0x`
Which was upgraded and stored in `tests/data/upgrades/sbm/v1_0`
Which was upgraded and stored in `tests/data/upgrades/sbm/v1_1`

So changes in the v0 model is the culprit.

LuukBlom: Also note that git doesnt show these additions to the file `tests/data/upgrade/sbm/v1_0/wflow_sbm.toml`.
That is because they were already there before in the sbm_with_lakes v0 model.

hydromt_wflow/wflow_sbm.py
LuukBlom: this change was done by Joost in b5d45c86e835fc5c1138fa788b8a5f1542450b6b

hydromt_wflow/version_upgrade.py
LuukBlom: Many of these checks are optional features for v1 that we upgrade manually in `_convert_to_wflow_v1`.
So they might not all be present in user models at the same time. 
To make sure we dont mis categorize, I think we should try to check and catch them all.

I agree with you, its not likely we ever get to this last `return True` statement, but that is just an order thing.
If we change the order, do we then remove a different one?

hydromt_wflow/version_upgrade.py
LuukBlom: For v0 to v1, I think starting from scratch is much simpler tbh. 
The structure of the toml file significantly changed, variables were moved, renamed etc.
Any config key we dont explicitly support in the version upgrade, is dropped, ensuring we can never accidentally forget to delete a v0 key, silently creating a mixed v0/v1 config.

For v1 to v1.1, the patch-style upgrade is already in place.

hydromt_wflow/version_upgrade.py
LuukBlom: toml libraries are for reading/writing to disk.
`get_config` and `set_config` work on already parsed toml files: in-memory dicts.
so Im not sure what we would gain by doing that.

hydromt_wflow/version_upgrade.py
LuukBlom: good catch. removed it now.

`_is_v1_schema` was only called by `_detect_version_from_config`, which checks the config for the `wflow_version` variable before calling `_is_v1_schema`, so this would not have bugged, but still good to be clear about intent.

```python

def _detect_version_from_config(config: dict) -> Version:
    """Detect the Wflow.jl version from a raw config dictionary.

    Parameters
    ----------
    config : dict
        The raw TOML config dictionary.

    Returns
    -------
    Version
        The detected Wflow.jl version.
    """
    version = get_config(config=config, key="wflow_version", fallback=None)
    if version is not None:
        logger.debug(f"Detected Wflow version {version} from config.")
        return Version(str(version))

    if _is_v1_schema(config):
        logger.debug(
            "No 'wflow_version' found in config but v1.0 schema detected. "
            "Assuming v1.0 model."
        )
        return Version("1.0")

    logger.warning(
        "No 'wflow_version' found in config and no v1.0-only keys detected. "
        "Assuming pre-v1.0 model with v0.8 schema. Please check the config and update "
        "the version if needed."
    )
    return Version("0.8")
```

hydromt_wflow/version_upgrade.py
```diff
@@ -861,10 +1152,342 @@ def convert_reservoirs_to_wflow_v1_sediment(
...
+    with open(DATA_DIR / "default_config_headers.toml", "rb") as file:
+        new_config = tomllib.load(file)
+
+    for key, value in config_out.items():
+        set_config(new_config, key, value)
+
+    # Write the upgraded config to disk
+    config_path = model_root / config_filename
+    write_toml(config_path, new_config)
+    logger.info(f"Wrote upgraded v1.0 config to {config_path}.")
```
LuukBlom, line 1389: Should we add a call to setup_basemaps + write here to update the staticmaps and config with the new required config var  land_surface__elevation`?

Or is the layer written in staticmaps.nc already? If so, we can just add this to the config and write the toml. No need to read a model to call setup_basemaps and write.
Also, is there a good default for `land_surface__elevation` @JoostBuitink

deltamarnix (commented):
Review comments incoming.

hydromt_wflow/version_upgrade.py
```diff
@@ -23,17 +25,468 @@
...
+    "lateral.river.h_av": None,
+    "lateral.river.floodplain.q": "floodplain_water__instantaneous_volume_flow_rate",
+    "lateral.river.floodplain.h": "floodplain_water__depth",
+    "lateral.river.lake.waterlevel": "reservoir_water_surface__elevation",
+    "lateral.river.reservoir.volume": None,
+    "vertical.paddy.h": "paddy_surface_water__depth",
+}
+
+# v0 variable to v1 variable for Sediment input variables
+_V0_TO_V1_INPUT_SEDIMENT: dict[str, str | None] = {
```
deltamarnix, line 197: This sediment v0→v1 mapping block is missing a few v0 aliases that are present in the test fixture and have explicit v1 targets: `vertical.slope` should map to `land_surface__slope`, `vertical.rivcell` to `river_location__mask`, `vertical.pclay`/`vertical.psilt` to the soil clay/silt fractions, and `lateral.river.fgravriv` appears misspelled here (`fgravriv` in the v0 TOML). As written, those layers will be dropped or stay unmapped during upgrade.

hydromt_wflow/naming.py
```diff
@@ -35,912 +35,715 @@
...
-        "wflow_v0": "lateral.land.h_av",
-        "wflow_v1": None,
+        "wflow_name": None,
     },
     "groundwater_head": {
-        "wflow_v0": "lateral.subsurface.flow.aquifer.head",
-        "wflow_v1": "subsurface_water__hydraulic_head",
+        "wflow_name": "subsurface_water__hydraulic_head",
     },
     "subsurface_q": {
```
deltamarnix, line 448: `subsurface_q` still points to `subsurface_water__volume_flow_rate`, but the v1.1 upgrade path and the current v1.1 fixture use `subsurface_water__instantaneous_volume_flow_rate`. Keeping the old name here means new models will advertise/export the pre-v1.1 state name instead of the latest schema.

hydromt_wflow/version_upgrade.py
```diff
@@ -861,10 +1152,441 @@ def convert_reservoirs_to_wflow_v1_sediment(
...
+    if ds_res is not None:
+        model.staticmaps.drop_vars(vars_to_remove)
+        model.staticmaps.set(ds_res)
+        for option in config_opt:
+            model.config.set(option, config_opt[option])
+    model.write()
+
+
+def _upgrade_components_v1_to_v1_1(
+    model_type: str,
```
deltamarnix, line 1475: I'm guessing this was already the case in the old upgrade functionality. But I think it would be nice to have an enum instead of a str defining how the flow of the code goes.

hydromt_wflow/version_upgrade.py
```diff
@@ -861,10 +1152,441 @@ def convert_reservoirs_to_wflow_v1_sediment(
...
+        ModelClass = _get_model_class(model_type)
+        model = ModelClass(
+            str(model_root),
+            config_filename=config_filename,
+            mode="r+",
+            data_libs=data_libs,
+        )
+        v0_opts = validated_options.get((Version("0.8"), Version("1.0")), {})
+        if model_type == "wflow_sbm":
+            _upgrade_components_v0_to_v1_sbm(model, config_v0, **v0_opts)
```
deltamarnix, line 1574: This code now still leans on config_v0, as it is copied and then passed to the upgrade functions to setup the reservoirs and tables.

I think it would be more understandable for developers if we would pass the updated v1 config and alter the update function so that it doesn't lean on the v0 names anymore, but on the v1 names. The config is updated, and then the tables and reservoirs are.

hydromt_wflow/version_upgrade.py
```diff
@@ -861,10 +1152,441 @@ def convert_reservoirs_to_wflow_v1_sediment(
...
+        elif model_type == "wflow_sediment":
+            _upgrade_components_v0_to_v1_sediment(model, config_v0, **v0_opts)
+
+    # Upgrade component data for v1 to v1.1 (needs initialized Model)
+    if needs_setup_basemaps:
```
deltamarnix: Wouldn't it be easier to check in the config if a certain variable is in or not? Then you don't need a return statement and the function becomes cleaner.

deltamarnix (changes requested):
We're heading in the right way. copilot did a thorough review for me on the naming to see if anything is still missing.

And on top of that I have some small code fixes that I think would be useful.

hydromt_wflow/version_upgrade.py
LuukBlom: `land_surface__slope` is already pointed to by `lateral.land.slope`, how should that be resolved?
Also, is there a list of renames done by wflow julia that made copilot find these things? Would be good to check them against each other.
This is part of the naming that Helene did, so im not super confident in changing these dicts.

For now I have added these changes.

hydromt_wflow/version_upgrade.py
LuukBlom: The component upgrade from v0 to v1 changes config structure and values.
To correctly upgrade the components as well, we need both the old config (describes what is currently in the staticmaps / component data) and the new (what should the new data look like after updating).

Unfortunately, I dont think its possible to upgrade the component data from just the data and the upgraded config since everything is user configurable.

deltamarnix: If you know exactly how the data was mapped to the new v1 format, you also know where to look for the info in the v1 config, right? Even when it has changed. What is the information that is missing in the v1 format? Everything should have been transferred and therefore findable.

LuukBlom: That is expected for a v0 to v1 upgrade though? 

Also, `_convert_sbm_reservoirs_v0_to_v1` takes a v0 config as an argument. 
So, perhaps it is possible to use the upgraded config, but that would require refactoring all of Helenes upgrade code, which I think is a bad idea since there are no tests.

deltamarnix: I agree, let's not refactor.

deltamarnix (approved):
Approving these changes. I think it is way nicer than before. Please create some issues on the things that you think should still be double checked. Like missing variables, and type-o's that you found in the old upgrade statements.

Issue #774, opened 2026-06-17
Support setup soil via vector input


michaelohanrahan:
#105 closes

But in the discussion: https://github.com/Deltares/hydromt_wflow/issues/105#issuecomment-2306639823

@hboisgon identifies this :
""""""
Idea for local ie vector soil data after discussions with @JoostBuitink.

Assumptions on available data:

a shapefile of points/polygons with soil/type
one/several csv with per soil profile the soil properties for different layer
image

Processing steps:

Create a raster version of the soil type (like for the landuse map). If the data is points, use IDW for interpolation or nearest, if polygons just rasterize.
Using a pandas Dataframe, compute each wflow parameter for each layer of each soil profile (should be able to re-use the functions in ptf.py as they are numpy based).
image
Then resample over the depths. After this step we should then get per soil type, the corresponding value of Ksat, M, thetaS, c_1, c_2 etc...
image
Map the wflow parameters values to the soil type map (same when we map landuse parameters to the landuse map)
Fill in the NaNs over the catchments

""""

PR #775, opened 2026-06-17
feat(soilgrids): support other/local soil data than SoilGrids (#105)

### LINKED ISSUE #105: Support other/local soil data than Soilgrids
We have two resampling method depending on parameter type (at depth or depth average).
We could introduce a switch and varying soil layer depths as input.

### LINKED ISSUE #772: soilgrids.py: Soil property information less than 2m not allowed
### Kind of request

### Enhancement Description

In some cases the deepest property layer of a soil is less than SoilGrids 2m max. If I have soil property layers that go to 1.5m those will be incompatible and averaging will dilute them. This hard-coded limit is a legacy of only using soil grids data where data is available up to 2m. Better advance soilgrids.py (also changing the name) to be less soil grids dependent. The most pragmatic initial solution being to extend the values of the soil property layer to the 2m horizon in a creating a homogeneous block to the max depth. The layering scheme can be attributes in the data_catalog or the dataset. 

### Use case

Using any layered soil data that doesnt exactly match either soilgrids (2017) or soilgrids_2020

### Additional Context

_No response_

michaelohanrahan:
Probably closes #105 but a separate issue is required for the discussion on vector soil setup report. #105 originally opened identifying the restrictions of using `soil_fn=="soilgrids_2020"` branching to identify block or trapezoidal layer averaging of soil properties. This limited the flexibility in using non soilgrids datasources. 

Simultaneously closes #772, where some sources require an averaging to 2m where the information in the source doesnt exist and is filled to the max required depth. 

## Issue addressed
Fixes #105 and #772

## Explanation
#105
1. ** Resampling switch by parameter type ** - introduce an input dataset attribute `soil_layer_type: point or interval` (accessible via `ds.attrs["soil_layer_type"]`) to switch between trapezoidal or block averaging, respectively. Removing `soilgrids_2020` branching, but for now retains fallback for legacy datasets that do not have the attribute. 

2. ** Varying soil layer depths as input ** - ideally sources self-describe via dataset attribute `soil_depth_edges_cm` combined with `soil_layer_type` so any n-layer source should work (similarly with fallback for legacy datasets)

#772
**solving soil average dilution** in the case that soil property layer limits <2m the soil property is assumed homogeneous from the last point of information to the base of the column. i.e. for a dataset with deepest property layer 60cm to 1m, the column to 2m is equal to the property values in that last layer. 

## Checklist
- [x ] Branch is up to date with `main`
- [] Updated documentation if needed

New issue required for #105 where vector support is identified moved to new issue #774

### DIFF
--- a/hydromt_wflow/workflows/soilgrids.py
+++ b/hydromt_wflow/workflows/soilgrids.py
@@ -19,22 +19,20 @@ __all__ = [
     "update_soil_with_paddy",
 ]
 
-# soilgrids_2017
-soildepth_cm = np.array([0.0, 5.0, 15.0, 30.0, 60.0, 100.0, 200.0])
-soildepth_mm = 10.0 * soildepth_cm
-# for 2017 - midpoint and midpoint surface are equal to original
-soildepth_cm_midpoint = np.array([0.0, 5.0, 15.0, 30.0, 60.0, 100.0, 200.0])
-soildepth_mm_midpoint = 10.0 * soildepth_cm
-soildepth_mm_midpoint_surface = 10.0 * soildepth_cm
+#Legacy: can be attached as attributes in the data catalog
+_SOIL_DEPTH_EDGES_CM = {
+    "soilgrids":      np.array([0.0, 5.0, 15.0, 30.0, 60.0, 100.0, 200.0]),
+    "soilgrids_2020": np.array([0.0, 5.0, 15.0, 30.0, 60.0, 100.0, 200.0]),
+}
+
+_SOIL_LAYER_TYPE = {
+    "soilgrids":"point",
+    "soilgrids_2020":"interval"
+}
+
+# Not depth-related; kept as live module constants.
 M_minmax = 10000.0
 nodata = -9999.0
-
-# default z [cm] of soil layers wflow_sbm: 10, 40, 120, > 120
-# mapping of wflow_sbm parameter soil_brooks_corey_c to soil layer SoilGrids
-c_sl_index = [2, 4, 6, 7]  # v2017 direct mapping
-# c_sl_index = [2, 3, 5, 6] #v2021 if direct mapping - not used,
-# averages are taken instead.
-
 
 def concat_layers(
     ds: xr.Dataset,
@@ -43,6 +41,11 @@ def concat_layers(
 ):
     """
     Preprocess functions to concat soilgrids along a layer dimension.
+
+    The number of layers is read from the data (``{var}_sl{i}`` variables) so any
+    source works regardless of ``soil_fn`` (7 for soilgrids 2017, 6 for
+    soilgrids_2020, 3 for the OpenLandMap-derived nc, ...) so long as the
+    convention is followed.
 
     Parameters
     ----------
@@ -53,10 +56,7 @@ def concat_layers(
     variables : list
         List of soil properties to concat.
     """
-    if soil_fn == "soilgrids_2020":
-        nb_sl = 6
-    else:
-        nb_sl = 7
+    nb_sl = _count_sl_layers(ds)
     ds = ds.assign_coords(sl=np.arange(1, nb_sl + 1))
 
     for var in variables:
@@ -76,7 +76,7 @@ def concat_layers(
     return ds
 
 
-def average_soillayers_block(ds, soilthickness):
+def average_soillayers_block(ds, soilthickness, soildepth_cm):
     """
     Determine weighted average of soil property at different depths over soil thickness.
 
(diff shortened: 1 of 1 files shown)


hydromt_wflow/workflows/soilgrids.py
```diff
@@ -450,6 +463,42 @@ def constrain_M(M, popt_0, M_minmax):
...
+        return np.asarray(ds.attrs["soil_depth_edges_cm"], dtype=float)
+    return np.asarray(_SOIL_DEPTH_EDGES_CM[soil_fn], dtype=float)
+
+
+def _layer_type(ds: xr.Dataset, soil_fn: str) -> str:
+    """'point' (values at depths, e.g. sg 2017) or 'interval' (averages between edges e.g. sg 2020)."""
+    lt = ds.attrs.get("soil_layer_type") or _SOIL_LAYER_TYPE.get(soil_fn)
+    if lt is None:  # neither attr nor registry -> infer from layer count
+        lt = "point" if _count_sl_layers(ds) == len(_calc_depth(ds, soil_fn)) else "interval"
+    return lt
```
LuukBlom, line 486: These functions deserve some testing and logging for debugging.
Could you add some debug statements and tests?

```python
def _layer_type(ds: xr.Dataset, soil_fn: str) -> str:
    """'point' (values at depths, e.g. sg 2017) or 'interval' (averages between edges e.g. sg 2020)."""
    layer_type = ds.attrs.get("soil_layer_type") or _SOIL_LAYER_TYPE.get(soil_fn)
    if layer_type is not None:
        logger.debug("layer type found in 'soil_layer_type' attribute or source is predefined.")
        return layer_type

    layer_count = _count_sl_layers(ds)
    layer_depths = _calc_depth(ds, soil_fn)
    layer_type = "point" if layer_count == len(layer_depths) else "interval"

    logger.debug(f"Neither attr nor registry -> inferred layertype '{layer_type}' from layer count: {layer_count} and depths: {layer_depth}")
    return layer_type
```

hydromt_wflow/workflows/soilgrids.py
LuukBlom, line 1: I like your additions here: small and obvious helper functions.

I also recommend adding type hints to all of the functions you create / touch, to make the code easier to read and self-documenting.

LuukBlom (commented):
Looking quite good already! There are some things that still need to happen as discussed, but a very good start, much more generic and backwards compatible than I was expecting.

The tests I'm mostly looking for:

- A test for `_layer_type` inference when neither `attr` nor registry match (the `lt is None` branch silently guesses `"point"` vs `"interval"` from layer count, without the user being notified). This needs to be (1) communicated, e.g. a warning or explicit error when inference is happening, and (2) tested, to make sure we don't silently return unexpected results.
- A test with a synthetic 3-layer source (e.g. mimicking OpenLandMap/SLGA).
- A test for the fix for #772 (soil data shallower than 2m).
- Tests for the other added private helper functions, where appropriate.

Quickstart for testing:

```python
def test_layer_type_from_attrs():
    # ds.attrs["soil_layer_type"] explicitly set -> respected, no inference

def test_layer_type_from_registry():
    # soil_fn matches legacy registry -> respected, no inference

def test_layer_type_inference_warns():
    # neither attrs nor registry match -> UserWarning / Error is raised, and asserts on
    # the specific inferred value so a silent behavior change gets caught

def test_layer_type_inference_ambiguous_raises():
    # layer count coincidentally matches len(depths) but source isn't
    # actually "point" data -> document expected behavior (warn vs raise)

def test_soilgrids_synthetic_3layer_source():
    # synthetic xr.Dataset mimicking OpenLandMap/SLGA (3 sl layers,
    # soil_depth_edges_cm + soil_layer_type set), run through soilgrids(),
    # assert shape/values of output maps

def test_average_soillayers_block_shallow_source():
    # test for issue #772
    # source max depth < wflow max depth -> edges[-1] extended to
    # max(soildepth[-1], wflow_depths[-1])

def test_average_soillayers_shallow_source():
    # test for issue #772
    # same as above but for the "point" averaging path, not just "_block"

def test_calc_depth():
def test_calc_midpoint():
def test_calc_surface():
def test_count_sl_layers():
    # small helper-level tests, mostly for edge cases (e.g. single-layer source)
```

Could you add tests along these lines in `tests/workflows/test_soildgrids.py`, including the synthetic datasets/fixtures needed to run them?

If you have any questions, please let me know!

LuukBlom:
Also, there is no vector based code here at all. 
I assume that will be added later, but this could already be merged if tested properly. 
Then we can do a second PR where we extend this system by adding the rasterization of the mosiac polygons.

For now, I will wait with more reviews.
If you want to have a design session before starting work on the rasterization, let me know, we can schedule something!

PR #777, opened 2026-06-18
Fix reservoirs outside rivers

ClaireDons:
## Issue addressed
Fixes #724 

## Explanation
Added ability for users to exclude reservoirs that are outside of the river network (no cells overlapping) by adding an exclude_outside_reservoirs flag, which is by default set to False. 

## Checklist


hydromt_wflow/workflows/reservoirs.py
LuukBlom, line 1: is the nodata value for reservoirs always -999 or is it user configurable? 

if so we should be more careful and use `ds.raster.nodata`

hydromt_wflow/workflows/reservoirs.py
```diff
@@ -65,11 +65,65 @@
...
+        reservoir_ids = reservoir_ids.where(
+            ~reservoir_ids.isin(outside_reservoirs), -999
+        )
+    elif not exclude_outside_reservoirs and len(outside_reservoirs) > 0:
+        logger.warning(
+            f"{len(outside_reservoirs)} reservoir(s), {outside_reservoirs.tolist()}, "
+            "are outside the river network. "
+            "Set 'exclude_outside_reservoirs=True' to exclude them."
+        )
+    return reservoir_ids
```
LuukBlom: We are scanning the the full raster for every reservoir in the list comprehension.
This can be simplified by computing the reservoid id + river mask intersection first.

Suggestion (still to change to use ds.raster.nodata if applicable?):
```python
    # Get unique reservoir IDs and exclude nodata value (-999)
    res_ids = np.unique(reservoir_ids.values)
    res_ids = res_ids[res_ids != -999]

    # Get reservoir IDs that overlap the river network
    river_res_ids = np.unique(
        reservoir_ids.where(river_mask == 1).values
    )
    river_res_ids = river_res_ids[river_res_ids != -999]

    # Reservoirs with no overlap with the river network
    outside_reservoirs = np.setdiff1d(res_ids, river_res_ids)

    if len(outside_reservoirs) > 0:
        if not exclude_outside_reservoirs:
            logger.warning(
                f"{len(outside_reservoirs)} reservoir(s), "
                f"{outside_reservoirs.tolist()}, are outside the river network. "
                "Set 'exclude_outside_reservoirs=True' to exclude them."
            )
            return reservoir_ids

        logger.warning(
            f"{len(outside_reservoirs)} reservoir(s), "
            f"{outside_reservoirs.tolist()}, were excluded because no cells "
            "were found within the river network. Consider decreasing the "
            "minimum upstream area threshold."
        )
        reservoir_ids = reservoir_ids.where(
                ~reservoir_ids.isin(outside_reservoirs), -999
        ) 
        return reservoir_ids

File: tests/test_model_methods.py [hunk 2]
```diff
@@ -276,8 +277,53 @@ def model_with_rating_curve_data(
...
+        "hydro_reservoirs",
+        geom=mod.region,
+        predicate="intersects",
+    )
+    extra = gdf_res.iloc[[0]].copy()
+    extra_id = int(gdf_res["waterbody_id"].max()) + 1
+    extra["waterbody_id"] = extra_id
+
+    river = mod.staticmaps.data["river_mask"]
+    x, y = 12.1, 46.45  # Hard coded coordinates outside the river network
```
LuukBlom: Could we avoid hardcoding the coordinates here? It would be more robust to derive a location from the model (e.g. pick a non-river cell from river_mask) so the test remains independent of the example model.

LuukBlom (changes requested):

hydromt_wflow/workflows/reservoirs.py
ClaireDons: It seems that -999 is currently assumed (although there is a function argument but it is otherwise hardcoded). I could open an issue to make it configurable?

LuukBlom: No, no need to make it configurable. 
I have seen bugs where in some places -999 was used and in others -9999, leading to bugs. 
So I think making it explicit by creating a variable and passing that around would be good.

tests/test_model_methods.py
ClaireDons: Ok I changed it now, I think not hardcoding it is much nicer

hydromt_wflow/workflows/reservoirs.py
ClaireDons: Ok I made the change in the functions I worked on and we can keep it in mind for other functions when we work on them

tests/test_model_methods.py
```diff
@@ -276,8 +278,75 @@ def model_with_rating_curve_data(
...
+    mask = (
+        (catchment > 0).values
+        & (river == 0).values
+        & (dist_river >= 0.03)
+        & (dist_border >= 0.05)
+    )
+
+    # Choose a random point and create a reservoir bbox
+    iy, ix = np.where(mask)
+    k = np.random.default_rng(100).integers(iy.size)
```
LuukBlom: Creating the random engine is actually quite expensive.

There is a session scoped fixture `rng` that creates this already and prevents doing this multiple times.

tests/test_model_methods.py
```diff
@@ -276,8 +278,75 @@ def model_with_rating_curve_data(
...
+    # Distance to nearest river cell & catchment border
+    dist_river = distance_transform_edt((river == 0).values, sampling=sampling)
+    dist_border = distance_transform_edt((catchment > 0).values, sampling=sampling)
+
+    # Mask of valid values
+    mask = (
+        (catchment > 0).values
+        & (river == 0).values
+        & (dist_river >= 0.03)
+        & (dist_border >= 0.05)
```
LuukBlom: Where do these numbers come from? Should we use the resolution?

tests/test_model_methods.py
```diff
@@ -324,6 +393,41 @@ def test_setup_reservoirs_no_control(
...
+        mod_with_extra.staticmaps.data["reservoir_area_id"].values
+    )
+    reservoir_warnings = [
+        r.message for r in caplog.records if r.name.endswith("workflows.reservoirs")
+    ]
+    assert any(
+        "were excluded because no cells were found within the river network." in msg
+        for msg in reservoir_warnings
+    )
+
```
LuukBlom: This test is getting a bit large and testing multiple things. Can you split this into multiple tests? 
Good rule of thumb is for each test to have at maximum 1 call to the function under test.

LuukBlom (commented):
Nice! 

Code LGTM, only very minor comments left about the tests

tests/test_model_methods.py
ClaireDons: I'll revert merging the tests then

tests/test_model_methods.py
ClaireDons: It's a distance that seemed sensible but I'll set them both to the same and use the raster cells since that actually simplifies the code. Here's a picture in case you are curious, the blue areas are where a point can be chosen, and red the reservoirs. 

image

LuukBlom (approved):

PR #785, opened 2026-06-30
Fix cells outside domain

### LINKED ISSUE #752: setup_basemap with a geometry results in staticmaps that is too big and contains cells outside model domain

### Reproducible Example

See my Bulletin for a reproducible example.

### Current behaviour

I have a DEM, from which I derived the hydrography data with HydroMT and PyFlwDir. Next, I want to build a Wflow model like so:

```python
import hydromt
from hydromt.model.processes.basin_mask import get_basin_geometry
from hydromt_wflow import WflowSbmModel

# Instantiate instance of Data Catalog
data_catalog = hydromt.DataCatalog(data_libs="path/to/data_catalog.yml")

# Read hydrography data
ds_hydrography = data_catalog.get_rasterdataset('hydrography')

# Read our basin index
basin_index = data_catalog["basin_index"]

# Outlet coordinates
x = my_x_coordinate
y = my_y_coordinate

gdf_bas, gdf_xy = get_basin_geometry(ds_hydrography,
                                     kind="subbasin",
                                     xy=[x, y],
                                     uparea=70) # set a minimum upstream area to force snapping to a river

# This creates a completely empty Wflow model.
model = WflowSbmModel("path/to/model",
                      data_libs="path/to/data_catalog.yml")

model.setup_basemaps(region={'geom': gdf_bas},
                     hydrography_fn='hydrography',
                     basin_index_fn='basin_index',
                     res=100) # resolution (unit is the same as the source CRS, meters in this case)

```

The resulting staticmaps is too big, and contains a couple of rows and columns outside the domain. Furthermore, the very last column is filled with values as well, instead of being NaN:

<details><summary>Image</summary>
<p>

Image

</p>
</details> 

### Desired behaviour

The staticmaps should be clipped as such to only contain the rows and columns needed to cover my model domain, and no more. It should certainly not have non-NaN data for cells outside my model domain.

ClaireDons:
## Issue addressed
Fixes #752 

## Explanation
The column filled with values instead of NaNs was traced back to nodata values outside of the domain being set to "True" when creating a mask for basins defined by "bbox" or "geom'. Now a mask is created by taking values above zero as active cells, which seems to also resolve the issue of the domain being too large too. 

## Checklist

It was difficult to pinpoint the issue due to the complexity of the hydrography function and its interactions with parse_region() and other clipping methods. We should consider some refactoring and clearer documentation for it in the future (#786).

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -16,6 +16,7 @@ Fixed
 -----
 - Fixed `setup_reservoirs_no_control()` to be able to handle both strings and path objects. (#770)
 - Fixed outdated imports in ``hydromt_wflow.components.staticmaps``. (#782)
+- Fixed `hydrography` workflow to correctly remove no data values outside of the model region. (#785)
 
 Removed
 -------
--- a/hydromt_wflow/workflows/basemaps.py
+++ b/hydromt_wflow/workflows/basemaps.py
@@ -169,17 +169,21 @@ parametrization of distributed hydrological models.
         # derive basins
         if np.any(outbas_pit != 0):
             idxs_pit = idxs_pit0[outbas_pit != 0]
-            basins = flwdir_out.basins(idxs=idxs_pit).astype(np.int32)
-            ds_out.coords["mask"] = xr.Variable(
-                dims=ds_out.raster.dims, data=basins != 0, attrs=dict(_FillValue=0)
-            )
+
         else:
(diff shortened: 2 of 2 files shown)


LuukBlom (approved):
LGTM,

Before merging, please put your findings for this issue in a comment in the issue, and perhaps create a new one for refactoring hydrography + adding tests

PR #796, opened 2026-07-09
Fix reservoir outlet and area

### LINKED ISSUE #182: Deriving the outlet and area of waterbodies
The method that is used to derive waterbodies outlet and area should be improved as it can lead to some wrongful delineation.
So far, we use:

1. Rasterize waterbody polygon using all_touched = True
2. Derive the outlet using the cell within the rasterized polygon with max upstream area value

Problem is that with this technique we do not really follow the flow directions so we can include cells in the mask that are part of a different subbasin/tributary. More problematic, in some cases the outlet of the waterbody falls in another basin. Example for the Danube near the sea outlet where a small lake is too close to Danube and the method above then puts the outlet of the lake in the Danube itself (I think this also happens for the Nile in the wetlands part):

image

Not too sure how to improve here but maybe all touched is a little radical to start with so maybe we should exclude cells from the area if the coverage of the cell by the lake is less than a certain threshold like 10%? (there's a nice raster.rasterize_geometry method in core to do that now too). Or used a different rasterization method than all_touched is True?
For the outlet, we should still make sure it falls under the right basin. The 10% filter might already help in some cases but we could also use the watershed area (Wshd_area) ie upstream area column of the waterbody database if they have it for better snapping. This type of snapping is also under dev in PR #172 for gauges.

ClaireDons:
## Issue addressed
Fixes #182 , #798 

## Explanation
I refactored reservoir_id_maps() into smaller functions for maintainability and added a configurable fraction parameter to exclude cells that have less than a certain threshold coverage of the cell by the lake (using the raster.rasterize_geometry from HydroMT Core as suggested). Requiring a minimum cell coverage of 10 % seems to solve the issue for that particular basin and other basins also seem to correspond better to the area. 

Screenshot 2026-07-09 155448

Screenshot 2026-07-09 155040

## Checklist

The default is set to None, which keeps the original behaviour of all_touched = True. Changing the rasterisation can change the location of outlets (which is intended), which can lead to some errors or duplicate IDs, when updating a model (merge_reservoirs())). Either merging is skipped or an error is thrown. It should be decided if we want to add an "overwrite" option or a updating the warning.


hydromt_wflow/workflows/reservoirs.py
```diff
@@ -216,33 +220,93 @@ def reservoir_id_maps(
...
+    for i in res_id:
+        res_acc = ds_like[uparea_name].where(ds_out["reservoir_area_id"] == i).load()
+        max_res_acc = res_acc.where(res_acc == res_acc.max(), drop=True).squeeze()
+        yacc = max_res_acc[ydim].values
+        xacc = max_res_acc[xdim].values
+        ds_out["reservoir_outlet_id"].loc[{f"{ydim}": yacc, f"{xdim}": xacc}] = i
+        outdf.loc[outdf.waterbody_id == i, "xout"] = xacc
+        outdf.loc[outdf.waterbody_id == i, "yout"] = yacc
+
+    return gpd.GeoDataFrame(outdf, geometry=gpd.points_from_xy(outdf.xout, outdf.yout))
```
JoerivanEngelen: Stylewise:

I know the xarray developers discourage accessing variables as attributes, not sure what the pandas people say, consider calling:

```suggestion
    return gpd.GeoDataFrame(outdf, geometry=gpd.points_from_xy(outdf["xout"] outdf["yout"]))
```

This at least clarifies that ``xout`` and ``yout`` isn't a fixed attribute of the Dataframe ``outdf``, but an dynamically added variable.

JoerivanEngelen (approved):
LGTM, added one minor annoyance about style in a comment

hydromt_wflow/workflows/reservoirs.py
ClaireDons: Good catch, I've updated elsewhere too now.

PR #803, opened 2026-07-16
Rebuild example models on windows and linux using gh actions

LuukBlom:
This workflow allows for the running of a pixi task and commiting its artifacts to any non-protected branch if there are any on the specified paths. (see inputs for all config options)

## Explanation

Currently, when things change, resulting in a difference in staticmaps, but we check and validate that the change is actually correct and expected, we need to update the example models / other artifacts on all supported platforms manually.

I made the inputs for the workflow configurable so we should be able to run different pixi tasks that regenerate other artifacts as well.

To test it, I added a push trigger with default values for all of the arguments, essentially the same as running with workflow dispatch and changing nothing.
For a successful run, see the action runs for this commit: [add testing code so push trigger works](https://github.com/Deltares/hydromt_wflow/pull/803/commits/9fcd9c95bf958de604060a69dba34a0c42f50101)

I have since removed the push trigger and the default param resolution, since we only want to run on request.

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -1,5 +1,5 @@
 ---
-name: Rebuild artifacts
+name: Build artifacts
 
 permissions:
   contents: write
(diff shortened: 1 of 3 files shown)


ClaireDons (approved):
LGTM :)

PR #811, opened 2026-07-21
Netcdf diff

deltamarnix:
Add automated install script for developers to add a git compare for netcdf files with the help of ncdump.
This makes it easier to understand the differences in a netcdf file without completely debugging it.

Simply call `pixi run install` and it will be installed along with the pre-commit hooks.
If a netcdf file has changed, you can call `git diff`, or perhaps your editor automatically supports that. VS Code had issues with this setup, while Git Fork seems to be understanding the git config that I gave it.

## Checklist


deltamarnix:
The result in git fork.

image

docs/dev_guide/dev_install.rst
```diff
@@ -21,6 +21,15 @@ once you are in the directory you can install the environment using pixi with
...
     $ pixi install
+    $ pixi run install
+
+The latter will install pre-commit hooks and a git compare tool for netcdf files.
+Comparing netcdf (.nc) files with git will be using ncdump.
+If you don't have this program installed on your computer, you can install it via pixi as such:
+
+.. code-block:: console
+
+    $ pixi global install --expose ncdump netcdf4
```
JoerivanEngelen, line 33: This requires a global install? Can't we add ncdump to the dev environment? Exposing netcdf4 globally seems a bit dangerous to me, given that there are also netcdf-related dependencies (e.g. xarray) in the dev environment?

JoerivanEngelen (approved):
Cool feature! I have one remark about the installation of ncdump, as I'm not a big fan of global installation.

docs/dev_guide/dev_install.rst
deltamarnix: If you install it in the development environment, you will need to `pixi run git diff`, because you need the project environment to be active. It's possible to do it that way, but my Git Fork instance, that I don't start with `pixi run fork`, will not find this environment. That's why I wanted ncdump in the path.

JoerivanEngelen: Ah yes I see, it also doesn't work if you do start your Git Fork instance in a ``pixi shell`` session?

deltamarnix: That will work, but makes it harder for me to start things. I will add some alternatives.

PR #816, opened 2026-07-24
Add upgrade CLI command and fix example notebooks

### LINKED ISSUE #807: Build docs dummy doesnt work

### Reproducible Example

broken docs-dummy: https://github.com/Deltares/hydromt_wflow/actions/runs/29561911369/job/87826013526
working full docs: https://github.com/Deltares/hydromt_wflow/actions/runs/29509233098/job/87658025003

### Current behaviour

Seems like in the examples, cli commands (`! hydromt build ...`) are not being executed.
This somehow is only a problem for docs-dummy and not the full docs pipeline.

### Desired behaviour

both work

### Additional context

_No response_

deltamarnix:
Still keep the hydromt cli calls for users to copy-paste onto their shell
Fixes the documentation
Added dedicated hydromt_wflow upgrade cli

## Summary

- Add standalone `hydromt_wflow upgrade` CLI command that upgrades wflow models without requiring a model instance or the `hydromt update` pipeline
- Remove `upgrade_to_latest` and `upgrade_to_v1_wflow` methods from model classes (unreleased, no backward compat needed)
- Update migration guide and example notebook to use the new CLI

## Motivation

The `upgrade_model()` function operates directly on files on disk, but the only way to invoke it was through `hydromt update`, which forces everything through `Model.update()` — reading the model into memory, swapping roots, and running steps. This conflicted with the file-based upgrade flow, especially when using `-o` for a separate output directory.

The new CLI calls `upgrade_model()` directly: `shutil.copytree` the source, then upgrade the copy in place.

## Changes

| File | Change |
|------|--------|
| `hydromt_wflow/cli.py` | New Click CLI with `upgrade` command |
| `pyproject.toml` | `console_scripts` entry point for `hydromt_wflow` |
| `hydromt_wflow/wflow_base.py` | Removed `upgrade_to_latest` and `upgrade_to_v1_wflow` |
| `tests/test_cli.py` | New CLI tests (help, validation, SBM/sediment upgrades, config parsing) |
| `tests/test_version_upgrade.py` | Removed `TestUpgradeToLatest` class |
| `examples/upgrade_to_wflow_v1.ipynb` | Updated to use new CLI, outputs to temp dir |
| `examples/data/wflow_upgrade/wflow_upgrade_sediment.yml` | Simplified upgrade config for sediment |
| `docs/user_guide/6_migration_guide/migration_wflow.rst` | Rewritten for new CLI |
| `docs/changelog.rst` | Added/Removed entries for CLI and old methods |

## CLI interface

```sh
hydromt_wflow upgrade ./my_model -o ./my_model_upgraded -v
hydromt_wflow upgrade ./my_model -o ./upgraded --model-type wflow_sediment -i config.yml -d artifact_data -v

### CODE REVIEW

File: examples/build_model.ipynb [hunk 1]
```diff
@@ -144,7 +144,11 @@
    "id": "15",
    "metadata": {},
    "source": [
-    "Once we have prepared our data catalog and model build settings file, we are ready to ask HydroMT to build our Wflow model. Building the model can be done from the command line. Down below is an example for the Northern part of the Piave basin in Italy:"
+    "Once we have prepared our data catalog and model build settings file, we are ready to ask HydroMT to build our Wflow model. Building the model can be done from the command line. Down below is an example for the Northern part of the Piave basin in Italy:\n",
```
ClaireDons: I was initially a little confused by the shell command being in the markdown and the "down below". The wording for the sediment model is a little clearer so maybe we can reuse that? 

```suggestion
    "Once we have prepared our data catalog and model build settings file, we are ready to ask HydroMT to build our Wflow model. Building the model can be done from the command line. Let's update the hydrological **Wflow** model available from our example for the Piave subbasin::\n",
```

examples/upgrade_to_wflow_v1.ipynb
```diff
@@ -37,7 +35,19 @@
...
+    "hydromt_wflow upgrade <v0_model_dir> -o <v1_model_dir> -v\n",
+    "```\n",
+    "\n",
+    "For a detailed overview of the upgrade cli options, run:\n",
+    "\n",
+    "```sh\n",
+    "hydromt_wflow upgrade --help\n",
+    "```\n",
+    "\n",
+    "Let's first look at the original TOML file before upgrading:"
```
ClaireDons, line 50: A couple of cells below we view the toml again, maybe we can delete one?

examples/upgrade_to_wflow_v1.ipynb
```diff
@@ -151,7 +165,7 @@
    "id": "14",
    "metadata": {},
    "source": [
-    "We can call the `upgrade_to_v1_wflow` function using the `update` CLI of HydroMT. Here is the configuration file that we will use:"
+    "For sediment models, you may need to pass upgrade options (e.g. which soil dataset to use). Create a simplified YAML config file:"
```
ClaireDons, line 168: I think I'm missing some context here. This seems to suggest we are creating a yaml file but we're just reading it. How was it created (or what does simplified mean in this case)? What is the difference between this yaml file and the "old" one a couple of cells below?

hydromt_wflow/cli.py
```diff
@@ -0,0 +1,117 @@
...
+    required=True,
+    type=click.Path(resolve_path=True, dir_okay=True, file_okay=False),
+    help="Output directory for the upgraded model. Must not already exist.",
+)
+@click.option(
+    "--model-type",
+    type=click.Choice(["wflow_sbm", "wflow_sediment"]),
+    default="wflow_sbm",
+    show_default=True,
+    help="Type of wflow model.",
```
ClaireDons: Could maybe give the choices here

examples/build_model.ipynb
ClaireDons: I realise now that the shell command being in the markdown but running the cli through python is done throughout, the wording here is a little unclear but the others are ok.

ClaireDons (approved):
Looks good, just a few comments that would make the examples a littler clearer (to me), but I'll approve in advance :) 

Btw, I also noticed that some of the links in the notebooks are broken again, even though I thought we had fixed them (and it doesn't look like they've been touched in this PR). Maybe we should create an issue to implement a link check for the notebooks?

examples/upgrade_to_wflow_v1.ipynb
LuukBlom: So to upgrade some models, it is required to run setup_basemaps / other setup_functions to correct staticmaps / other data.

These functions have input args (input soild dataset etc) that need to be configurable for users when upgrading.

This format for this yml is owned and validated by the `hydromt_wflow.version_upgrade.py` module.

I put the upgrade yml in the v0 dir since its needed to upgrade that model.

LuukBlom: Since its possible that different version upgrades require different args, the format is like this:
```yml
<from_version>_<to_version>: { <setup_method_name>: { **kwargs }  },
...

```

PR #831, opened 2026-08-17
feat: set config entry for 1d floodplains

### LINKED ISSUE #762: Estimate floodplain slope
### Kind of request

Adding new functionality

### Enhancement Description

As part of upcoming release Wflow v1.1.0 kinematic wave routing supports a 1D floodplain schematization (already in place for local inertial routing).

```toml
[input.static]
floodplain__slope = "FloodplainSlope"
``` 

This way different slope values can be used for the river and floodplain flow routing.

### Use case

_No response_

### Additional Context

_No response_

LuukBlom:
## Issue addressed
Fixes #762

## Explanation
As discussed with @JoostBuitink and @DirkEilander, for now we are just setting the floodplain slope variable to be the same as the river slope.

This PR also updates pyflwdir to the latest release, since there have been significant bugfixes there since the current minimum version

### DIFF
--- a/hydromt_wflow/wflow_sbm.py
+++ b/hydromt_wflow/wflow_sbm.py
@@ -10,7 +10,6 @@ import geopandas as gpd
 import hydromt
 import numpy as np
 import pandas as pd
-import pyflwdir
 import xarray as xr
 from hydromt import hydromt_step
 from hydromt.error import NoDataException, NoDataStrategy
@@ -373,6 +372,10 @@ class WflowSbmModel(WflowBaseModel):
         * **river_bank_elevation** map: hydrologically conditioned elevation [m+REF]
           (for 2D floodplains)
 
+        For 1D floodplains, the ``floodplain__slope`` config entry is also set to
+        point to the river slope map, as the floodplain slope is for now assumed to
+        be equal to the river slope.
+
         Required setup methods:
 
         * :py:meth:`~WflowSbmModel.setup_rivers`
@@ -412,8 +415,8 @@ class WflowSbmModel(WflowBaseModel):
         """
         if self.config.get_value("model.river_routing") != "local_inertial":
             raise ValueError(
-                "Floodplains (1d or 2d) are currently only supported with \
-local inertial river routing"
+                "Floodplains (1d or 2d) are currently only supported with "
+                "local inertial river routing"
             )
         # update self._MAPS and self._WFLOW_NAMES with user defined output names
         var = "floodplain_water__sum_of_volume_per_depth"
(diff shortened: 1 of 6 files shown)


hydromt_wflow/wflow_sbm.py
```diff
@@ -412,8 +415,8 @@ def setup_floodplains(
         """
         if self.config.get_value("model.river_routing") != "local_inertial":
             raise ValueError(
-                "Floodplains (1d or 2d) are currently only supported with \
-local inertial river routing"
```
JoerivanEngelen, line 416: Tip: You can use ``textwrap.dedent`` to create multiline messages.

JoerivanEngelen (approved):
Looks good to me, I've got one tip to deal with multiline messages slightly nicer:

```python
import textwrap

def some_func():
    msg = textwrap.dedent(
        """
        This is a multiline message.
        It can be indented nicely in the code, but will be left-aligned when printed.
        """
    )

    raise ValueError(msg)
```

hydromt_wflow/wflow_sbm.py
LuukBlom: Yeah in some cases I like to use that as well! 
In this case that increases the indentation another time, reducing the usable line length even further.

This also works, and python concats / adds the strings in the same way:
```python
msg = (
    "Floodplains (1d or 2d) are currently only supported with "
    "local inertial river routing"
)
print(msg)
```
output:
```text
Floodplains (1d or 2d) are currently only supported with local inertial river routing
```

hydromt_wflow/wflow_sbm.py
```diff
@@ -412,8 +415,8 @@ def setup_floodplains(
...
+                "Floodplains (1d or 2d) are currently only supported with "
+                "local inertial river routing"
```
JoostBuitink, line 419: This error is no longer 100% correct.

`floodplain_1d__flag` is now allowed in wflow with the following river_routing options `local_inertial` (same as before), `kinematic_wave` and `manning_staggered`. Note that the last option is fully new.

2d floodplains are still only supported with river_routing = "local_inertial"

hydromt_wflow/wflow_sbm.py
```diff
@@ -373,6 +372,10 @@ def setup_floodplains(
         * **river_bank_elevation** map: hydrologically conditioned elevation [m+REF]
           (for 2D floodplains)
 
+        For 1D floodplains, the ``floodplain__slope`` config entry is also set to
+        point to the river slope map, as the floodplain slope is for now assumed to
+        be equal to the river slope.
+
```
JoostBuitink, line 378: Good to mention here that this entry is only actually required for when `river_routing = "kinematic_wave"`. No need to change the logic I'd say, and we just always add the entry if 1D floodplains are required (for easy switching if that is required)

PR #844, opened 2026-09-07
filter landuse for agroforestry

### LINKED ISSUE #841: setup_agroforestry convert specific land use types within provided polygon
### Kind of request

Adding new functionality

### Enhancement Description

At this moment, it is not yet possible to convert specific land use types within a provided polygon to agroforestry. This would be a useful additional feature, especially in the development of the QGIS plugin. 

It would be good to add an optional argument vector_fn (or something similar) in the current setup_agroforestry which allows to convert specific land use types to agroforestry within this polygon.

### Use case

_No response_

### Additional Context

_No response_

hboisgon:
## Issue addressed
Fixes #841 

## Explanation
Added a ``landuse_class_filter`` argument to list which landuse type to convert to agroforestry within zones defined in ``agroforestry_fn``. Updated the docs with an example and added a figure with the outputs.

## Checklist


laurenebouaziz (approved):
This is exactly the required additional improvement which was required, thanks @hboisgon !

hydromt_wflow/workflows/landuse.py
```diff
@@ -571,6 +571,7 @@ def add_agroforestry_to_landuse(
     ds_like: xr.Dataset,
     agroforestry_class: int | None,
     output_agroforestry_class: int | None = None,
+    landuse_class_filter: list[int] = [],
```
ClaireDons: Since lists are mutable, I would consider changing this to: 

```suggestion
    landuse_class_filter: list[int] | None = None,
```
and then having a:
```
if landuse_class_filter is None:
    landuse_class_filter = []
```

within the method to make sure that a new list is created every time it is called. Although this pattern is elsewhere in the codebase so it might be ok in this case.

hydromt_wflow/wflow_sbm.py
```diff
@@ -1537,6 +1537,7 @@ def setup_agroforestry(
         agroforestry_fn: str | xr.DataArray | gpd.GeoDataFrame,
         agroforestry_class: int | None = None,
         output_agroforestry_class: int | None = None,
+        landuse_class_filter: list[int] = [],
```
ClaireDons: Same here and then initialise in a similar manner to `lulc_mix_classes`

ClaireDons (commented):
Hey, looks good to me! Just a couple of minor comments that might help with maintainability or avoiding any strange bugs in the future. 

I'm holding off on approving for now as Ali and John need to make a decision about whether we can merge the Nature Based Solutions related work into main.

hboisgon:
Thanks @ClaireDons ! I updated with your comments.
I thought that merging was only delayed for the ponding methods. This adds a new argument to a function that is already in main (agroforestry). Do we really need to delay the merge?

ClaireDons:
> Thanks @ClaireDons ! I updated with your comments. I thought that merging was only delayed for the ponding methods. This adds a new argument to a function that is already in main (agroforestry). Do we really need to delay the merge?

Ah sorry, I probably got confused, if it's already implemented by Wflow then it is fine to merge into main!

